PR #25, opened 2016-11-28
add an aion_inv to network_module and use this to replace divides

zingale:
this is tested here:

http://bender.astro.sunysb.edu/Microphysics/test-suite/test-suite-gfortran/2016-11-28/index.html

only 1 test has high errors, test_react-ctest -- not sure what the issue is there, but the changes are minimal.

### DIFF
--- a/EOS/gamma_law_general/actual_eos.F90
+++ b/EOS/gamma_law_general/actual_eos.F90
@@ -12,7 +12,7 @@ module actual_eos_module
   use bl_space
   use bl_error_module
   use bl_constants_module
-  use network, only: nspec, aion, zion
+  use network, only: nspec, aion, aion_inv, zion
   use eos_type_module
 
   implicit none
@@ -70,7 +70,7 @@ contains
     if (assume_neutral) then
        state % mu = state % abar
     else
-       state % mu = ONE / sum( (ONE + zion(:)) * state % xn(:) / aion(:) )
+       state % mu = ONE / sum( (ONE + zion(:)) * state % xn(:) * aion_inv(:) )
     endif
     
     !-------------------------------------------------------------------------
--- a/integration/BS/bs_jac.F90
+++ b/integration/BS/bs_jac.F90
@@ -8,8 +8,7 @@ contains
 
     !$acc routine seq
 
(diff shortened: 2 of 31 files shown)


dwillcox:
Is this the right link to the test? On this link, I see a lot of failures and test_react-ctest did not compile.

zingale:
http://bender.astro.sunysb.edu/Microphysics/test-suite/test-suite-gfortran/2016-11-28-001/index.html

dwillcox:
Thanks! I'm not sure what is going wrong with test_react-ctest ... 

It only has the reaction C12 + He4 -> O16 though and the integration time is long enough to arrive at X(C12) < X_small. I suspect that could make the output more sensitive to the additional roundoff.

dwillcox:
But I looked at all the source changes, and I agree they are minimal and, beyond roundoff, shouldn't make a difference.

dwillcox:
Following up, if I change `inputs_reaclib_ctest` so that the integration time is only 1.0e-6 seconds instead of 1.0e-1 seconds, then fcompare gives the following between the `test_react` outputs for development and aioninv (before the merge):

```
            variable name            absolute error            relative error
                                        (||A - B||)         (||A - B||/||A||)
 ----------------------------------------------------------------------
 level =  1
 density                            0.000000000               0.000000000    
 temperature                        0.000000000               0.000000000    
 Xnew_helium-4                      0.000000000               0.000000000    
 Xnew_carbon-12                     0.000000000               0.000000000    
 Xnew_oxygen-16                     0.000000000               0.000000000    
 Xold_helium-4                      0.000000000               0.000000000    
 Xold_carbon-12                     0.000000000               0.000000000    
 Xold_oxygen-16                     0.000000000               0.000000000    
 wdot_helium-4                      0.000000000               0.000000000    
 wdot_carbon-12                     0.000000000               0.000000000    
 wdot_oxygen-16                     0.000000000               0.000000000    
 rho_Hnuc                           0.000000000               0.000000000    
 PLOTFILES AGREE
```

Which supports my previous assertion that the difference for the ctest network is its vulnerability to roundoff accumulation once X_C12 < X_small. In any case, the ctest results do not indicate a problem with the aioninv changes.

PR #34, opened 2017-06-08
fix from Aron Michel (HITS) that fills in missing reactions.

zingale:
In particular, the additional terms for 56Fe and 56Cr are now added.  

From Aron:

```
I have isolated the fixes for the RHS of the approx21 network. The
missing terms are the reactions to 56Fe and 56Cr. The latter one
represents the neutron rich material in the approx21 network. Therefore
neutron enrichment does not take place in the Boxlib version of the
network.

The additional terms for the reactions to 56Fe and 56Cr have been
obtained by comparing the approx21 network in Boxlib with the one
provided by F. Timmes
```


zingale:
thanks for the contribution!

PR #39, opened 2017-08-11
Move retry functionality up to main integrator

maxpkatz:
By doing a retry in the main integrator instead of the individual integrators, this allows us to combine multiple integrators in a single burn. The value is that if a burn fails, we can try a different integrator. Previously, our only recourse in that case was to loosen the tolerance on the burn and try again. We still will loosen the tolerances if necessary, but not before trying multiple integrators and seeing if one of them can do the burn.

At present only VODE90 and BS can be combined this way. VBDF and VODE90 continue to use the old method.


maxpkatz:
The order in which we compile in the integrator is a little brittle, it affects which version of rpar.F90 gets pulled in. We should fix that.

maxpkatz:
We should add a test that exercises the retry_burn functionality.

maxpkatz:
This is ready for review and a Castro test suite run.

zingale:
this passes all tests

http://bender.astro.sunysb.edu/Castro/test-suite/test-suite-amrex-gfortran/2017-08-14-001/index.html

PR #64, opened 2017-10-19
add the ability to boost the reaction rate

zingale:
add the ability to boost the reaction rates

### DIFF
--- a/integration/BS/bs_jac.F90
+++ b/integration/BS/bs_jac.F90
@@ -14,7 +14,7 @@ contains
     use actual_rhs_module, only: actual_jac
     use numerical_jac_module, only: numerical_jac
     use extern_probin_module, only: jacobian, burning_mode, burning_mode_factor, &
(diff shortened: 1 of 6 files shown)


zingale:
waiting for comments before I change the other integrators

dwillcox:
You might not be using a numerical jacobian, but if you do, its calls to `actual_rhs` should be scaled.

dwillcox:
This looks reasonable to me though.

maxpkatz:
Why only boost by a positive factor? One might want to slow the reactions down too. Test on react_boost > 0.

zingale:
I'd done the rest of the burners (but not SDC yet), and I've address Max's comment about react_boost > 0.

This is ready to merge.  See: https://github.com/starkiller-astro/Microphysics/blob/development/Docs/tests/ode_scaling.ipynb for a demonstration of this method with VODE.

PR #65, opened 2017-10-21
remove esum() for sums of 1 or 2 terms

zingale:
`esum()` has no benefit for sums of 1 or 2 terms


zingale:
test suite run of this is here:
http://bender.astro.sunysb.edu/Microphysics/test-suite/test-suite-gfortran/2017-10-21/index.html

(the aprox13 failures there are a filename clash)

PR #68, opened 2017-10-22
move the internal EOS functions to pure functions

zingale:
this marks all of the internal table functions as pure, in hopes that some compilers might find that info useful.


zingale:
unit test with `test_eos` shows no differences with this change.  Timings are as follows:

old: 0m2.135s 0m2.251s 0m2.215s 0m1.927s 0m2.005s 0m2.139s                      

new: 0m1.829s 0m1.901s 0m2.019s 0m2.281s 0m2.122s 0m1.839s                      

There's a lot of variability, so it is hard to conclude whether it is faster for real.

PR #88, opened 2018-03-15
Do not define extra thermo components of eos_t ifndef EXTRA_THERMO

dwillcox:
This PR will not declare composition derivative components of `eos_t` unless `EXTRA_THERMO` is defined.

Previously, we would not yield a compiler error if we tried accessing the composition derivatives without `EXTRA_THERMO`, introducing undefined results and, when compiling with OpenMP, silent NaNs.

We should use the `eos_t.F90` here in Maestro to avoid problems like this.

This changes one instance of `if EXTRA_THERMO` to `ifdef EXTRA_THERMO` in `helmholtz/actual_eos.F90` to conform to all other ifdef statements in which `EXTRA_THERMO` appears.

### DIFF
--- a/EOS/helmholtz/actual_eos.F90
+++ b/EOS/helmholtz/actual_eos.F90
@@ -219,7 +219,7 @@ contains
                             detadt,detadd,xnefer,dxnedt,dxnedd,s, &
                             temp,den,abar,zbar,ytot1,ye
 
-#if EXTRA_THERMO
+#ifdef EXTRA_THERMO
         !..for the abar derivatives
         double precision :: dpradda,deradda,dsradda, &
                             dpionda,deionda,dsionda, &
--- a/GMicrophysics.mak
+++ b/GMicrophysics.mak
@@ -29,7 +29,7 @@ endif
 EOS_TOP_DIR := $(MICROPHYSICS_HOME)/EOS
 
 ifdef EXTRA_THERMO
-  ifeq($(EXTRA_THERMO), t)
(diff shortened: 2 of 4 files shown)


dwillcox:
So should I just merge this?

maxpkatz:
I'm confused why the dpdX thing isn't causing a problem in Castro.

dwillcox:
I'm looking at the Castro changelog and for 17.11:

>  -- we no longer get the compositional derivatives as part of the EOS                                                                                                                                                                       
>     call.  If you need this functionality, you need to set the                                                                                                                                                                              
>     preprocessor variable (in Fortran), EXTRA_THERMO

maxpkatz:
Oh, because Castro has its own copy eos_type.f90.

dwillcox:
In Castro's copy of eos_type.F90, those compositional derivatives are only part of eos_t if EXTRA_THERMO is defined.

dwillcox:
Which is the same thing this PR does to the eos_type interface in Microphysics.

zingale:
fine with me -- want me to run the test suite on it?

maxpkatz:
Yes.

zingale:
There are a lot of failures with this -- seems the pressure is wrong now?http://bender.astro.sunysb.edu/Microphysics/test-suite/test-suite-gfortran/

dwillcox:
The conflict is resolved.

dwillcox:
Thanks!

PR #94, opened 2018-04-04
Updated subch network to include the C14(a,g)O18 reaction.

abigailbishop:


dwillcox:
I noticed `networks/subch/20180319default2` is part of the PR. 

I am in favor of keeping the Reaclib snapshot files we use to generate networks in the network directories in Microphysics for reproducibility. Because we decided not to include library snapshots in pynucastro though I wanted to point it out and discuss @zingale

zingale:
that's fine with me too

maxpkatz:
Are we going to be checking in a lot of multi-hundred-thousand-line files, or is this a one-off thing?

zingale:
that's a fair point.  Since this is the first of the automagically generated nets, it is kinda an experiment.
Perhaps for later ones, we can store the URL that can get the reaclib database used -- I think you can always get the older ones

maxpkatz:
If the URL to the database is stable and we say what version of the code we used to generate the net and include instructions for how to do it, I think that is sufficient to count as reproducible.

PR #96, opened 2018-04-06
Modify burn retry strategy to stick with current integrator as long as possible

### LINKED ISSUE #95: Change retry to loosen tolerances first, then try changing integrators
As per discussion in Slack, we'd like to change the retry functionality to first loosen the tolerances if an integration fails, and if it still fails after that we can try a different integrator.

maxpkatz:
Instead of swapping integrators at a fixed tolerance, we will instead hold the integrator constant and first attempt loosening the tolerance when a burn fails. This allows for a more consistent evolution when a burn fails, at the expense of a looser tolerance.

Closes #95

### DIFF
--- a/integration/integrator.F90
+++ b/integration/integrator.F90
@@ -118,7 +118,13 @@ contains
 #elif (INTEGRATOR == 1)
                 print *, "Retrying burn with VODE integrator"
 #endif
-             endif
+             else
+
+                ! We are out of integrators; give up.
+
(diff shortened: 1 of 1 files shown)


zingale:
I don't see how we exit that inner `do` loop if the retry fails, and then try the other integrator.  Don't we need an `exit` somewhere?

maxpkatz:
Fixed.

PR #100, opened 2018-04-12
Fix the calculation of the numerical Jacobian terms wrt X

### LINKED ISSUE #92: numerical jacobian may not be doing the species term correctly?
We seem to be computing d(Ydot)/dX in `numerical_jacobian.f90` -- it is not clear we should be mixing mass and molar fractions here.

maxpkatz:
These were missing a factor of A, so the Jacobian was actually using d(Y-dot) / dX.

Fixes #92


zingale:
I'm going to run the suite, since answers will change due to solver differences

PR #108, opened 2018-05-16
add burn_type module so this compiles

ajnonaka:


zingale:
can you leave in the `implicit none`?

Also, the approach that we take with both Maestro and Castro is that they carry their own version of `general_null` so one can compile those application codes without `Microphysics` if desired.  Just to lower the dependencies needed to get started.

ajnonaka:
The implicit none is in each subroutine, as opposed to at the top of the module (so functionally identical at the moment) --- but ---

I actually copied in the version in MAESTROeX and CASTRO, so do we want all 3 versions to be the same?

This all came about because for some reason Rupert wanted to build in the Microphysics version of general_null instead of MAESTROeX's version.  I suggested to Rupert he just use the MAESTROeX version, but he did note that the Microphysics version should actually compile anyway.

maxpkatz:
Yeah they should all be the same.

zingale:
closing since this issue is old, and we haven't seemed to need this for our runs

PR #124, opened 2018-06-18
Update interfaces for CUDA.

dwillcox:
This updates the burn type and eos type in /interfaces for CUDA managed memory and adds AMREX_DEVICE to the appropriate subroutines.

Allocating and initializing the eos module variables is now in unit_test/eos.F90.


zingale:
all regression tests passed

PR #125, opened 2018-06-18
Update neutrinos module for CUDA

dwillcox:
This updates the neutrinos module for CUDA. Mostly just adding AMREX_DEVICE.


zingale:
all tests pass

PR #126, opened 2018-06-18
Update screening module for CUDA

dwillcox:
This updates the screening module to use CUDA managed variables and adds AMREX_DEVICE.


zingale:
all tests passed

PR #127, opened 2018-06-18
Markup EOS routines for CUDA

dwillcox:
This adds CUDA markup to the EOS subroutines, except for stellarcollapse and ztwd.


zingale:
all tests pass

PR #128, opened 2018-06-18
Add utility modules and specialized esums

dwillcox:
This adds the CUDA-marked up utility modules for BLAS and LINPACK from cudadevice.

This also adds the specialized esums to microphysics math and replaces `esum` in the aprox networks with the specialized versions.


zingale:
all tests passed

PR #129, opened 2018-06-18
Mark up rates module used by aprox networks for CUDA

dwillcox:
This marks up the rates module subroutines for CUDA.


zingale:
all tests pass

PR #130, opened 2018-06-18
Use the CUDA-enabled reactions test driver.

dwillcox:
This switches to the test react driver from the `cudadevice` branch.

No test results should change as a result.


unit_test/test_react/main.F90
```diff
@@ -159,101 +185,72 @@ program test_react
...
 
-  ! note: integer division
-  n_rhs_avg = n_rhs_avg/(nT*nrho*nX)
-
-  print *, "RHS stats:"
-  print *, "  min: ", n_rhs_min
-  print *, "  avg: ", n_rhs_avg
-  print *, "  max: ", n_rhs_max
-
   ! output
```
zingale: does this PR eliminate the stats (min, avg, max)?  Can we keep those in for CPU?

zingale:
all tests pass

PR #133, opened 2018-06-20
Add CUDA version of VODE90 to the integration options

dwillcox:
This adds the CUDA-enabled VODE90 integrator from the cudadevice branch in place of the existing VODE90 integrator.

Minor modifications to the integration driver as well.


integration/VODE90/actual_integrator_sdc.f90
```diff
@@ -3,9 +3,9 @@
 
 module actual_integrator_module
 
-  use amrex_error_module
-  use amrex_constants_module
-  use amrex_fort_module, only : rt => amrex_real
+  use bl_error_module
+  use bl_types
+  use bl_constants_module
```
zingale: why are we using bl_error_module instead of amrex_error_module?

integration/VODE90/actual_integrator_sdc.f90
```diff
@@ -72,19 +72,19 @@ subroutine actual_integrator(state_in, state_out, dt, time)
 
     type (sdc_t), intent(in   ) :: state_in
     type (sdc_t), intent(inout) :: state_out
-    real(rt),    intent(in   ) :: dt, time
+    real(dp_t),    intent(in   ) :: dt, time
```
zingale: shouldn't we be using `rt` like we were before?

integration/VODE90/vode_rhs_sdc.f90
```diff
@@ -6,9 +6,9 @@
 subroutine f_rhs(neq, time, y, ydot, rpar, ipar)
 
   use actual_network, only: aion, nspec_evolve
+  use bl_types, only: dp_t
```
zingale: aren't we using `rt`?

integration/VODE90/actual_integrator_sdc.f90
dwillcox: I just fixed this to use the amrex modules.

integration/VODE90/actual_integrator_sdc.f90
dwillcox: yes, I switched dp_t to rt using amrex_fort_module

integration/VODE90/vode_rhs_sdc.f90
dwillcox: yep!

PR #145, opened 2018-08-13
Various optimizations to the Helmholtz EOS

maxpkatz:


maxpkatz:
Ready for review. Main changes were:
- Cleaning out some dead code paths (compiler may have already been optimizing these out)
- Replacing a number of divides with multiplications
- Combining some of the EOS tables for improved memory access patterns

maxpkatz:
For the wdmerger 2D collision test, this yields about a 3% performance improvement for a run with hydro and gravity.

maxpkatz:
In my testing, the diffs for pressure and temperature are a hard zero, but the velocities have a small but non-zero difference. Not sure if that is an error or just numerical roundoff.

zingale:
I'll run the suite on this

dwillcox:
From Max:

>#145 is bad and I don't want to merge it

PR #164, opened 2018-10-27
Add conductivity to eos_type when compiling with USE_EXTRA_THERMO=TRUE

dwillcox:
Unify the conductivity interface with the eos interface style by putting conductivity into the eos type if `USE_EXTRA_THERMO=TRUE`.

This gets rid of temporary variable copies in `actual_conductivity` and allows us to call `actual_conductivity` from Python via `python_library`.

Also this updates variable declaration style in the conductivity module and adds !$gpu.


dwillcox:
Related to the following PRs in MAESTRO and MAESTROeX:

https://github.com/AMReX-Astro/MAESTRO/pull/25

https://github.com/AMReX-Astro/MAESTROeX/pull/23

zingale:
Castro will also need the change

dwillcox:
Alright, this puts conductivity into the eos type but only compiles in the conductivity routines if `USE_CONDUCTIVITY=TRUE` for `Make.Microphysics`.

dwillcox:
Related -- https://github.com/AMReX-Astro/Castro/pull/431

dwillcox:
This PR adds `unit_test/test_conductivity` but for the purposes of testing this PR I could mimic that test with the old conductivity interface in development.

Make.Microphysics
```diff
@@ -178,10 +188,10 @@ ifeq ($(findstring gamma_law_general, $(EOS_DIR)), gamma_law_general)
    DEFINES += -DEOS_GAMMA_LAW_GENERAL
 endif
 
-# include the conductivity if not compiling for GPU
-# conductivity has not been ported/tested with CUDA Fortran
-ifdef USE_CUDA
-  ifneq ($(USE_CUDA),TRUE)
+# include the conductivity
+ifdef USE_CONDUCTIVITY
```
zingale: is this line needed?

dwillcox: Castro and Maestro need the conductivity, but not, say, test_react or test_eos in Microphysics itself.

For those cases we don't need to compile the conductivity in.

If you would prefer to compile conductivity in even for those unit tests then I can take this ifdef out.

zingale: oh now.  I see.  That's fine

dwillcox:
Okay, I just rebased onto development and force-pushed to this branch.

PR #170, opened 2019-02-25
ECSN

sailoridy:


zingale:
can you add a short description about the purpose of these changes to this github conversation so we can have the commit message be descriptive

sailoridy:
This commit is to calculate a one-cell burn by using the ECSN networks

maxpkatz:
Can you please change the target branch to `development` and then make sure your changes are in sync with that, if necessary?

dwillcox:
This was merged in #173

PR #190, opened 2019-09-19
Update Nonaka plot feature

dwillcox:
This updates the support for making Nonaka plots for all the integrators.

Along the way, this PR rearranges the modules in `integration/` for clarity.

Some utility modules useful for all integrators but not directly called by user code has been moved to `integration/utils`.


integration/BS/bs_jac_sdc.F90
```diff
@@ -32,7 +32,7 @@ subroutine bs_jac(bs)
 
        call bs_to_burn(bs, burn)
 
-       call actual_jac(burn)
+       call network_jac(burn)
```
zingale: why all the changes from `actual_jac` to `network_jac`?

dwillcox: This doesn't yet serve a purpose, but I think it would be interesting to add Nonaka-type plots showing the state where we evaluate the Jacobian.

PR #191, opened 2019-09-25
Add attributes(device) if not compiled with AMREX_USE_GPU_PRAGMA for …

jmsexton03:
…linear_algebra_modules


dwillcox:
Thanks @jmsexton03 !

PR #194, opened 2019-09-27
VODE90 for simplified SDC

zingale:
These changes let us use the VODE90 integrator for the Simplified-SDC


zingale:
this is ready for review

tests do as they should

dwillcox:
Since we're changing `USE_SDC` to `USE_SIMPLIFIED_SDC` can we also change the preprocessor definition from `SDC` to `SIMPLIFIED_SDC`?

That would clarify what's meant by:

```
#ifdef SDC
 ...
#elif TRUE_SDC
 ...
#else
 ...
#endif
```

Otherwise looks good!

zingale:
I think this is ready for re-review

Note: because of this change, any application codes will need to switch their builds to set `-DSIMPLIFIED_SDC` instead of `-DSDC`

integration/VODE90/actual_integrator_simplified_sdc.F90
```diff
@@ -97,40 +60,46 @@ subroutine actual_integrator(state_in, state_out, dt, time)
...
     !
     ! **NOTE** if you reduce these tolerances, you probably will need
     ! to (a) decrease dT_crit, (b) increase the maximum number of
     ! steps allowed.
 
-    atol(SFS:SFS-1+nspec) = atol_spec ! mass fractions
-    atol(SEDEN)           = atol_enuc ! total energy
-    atol(SEINT)           = atol_enuc ! internal energy
+    dvode_state % atol(SFS:SFS-1+nspec) = atol_spec ! mass fractions
+    dvode_state % atol(SEDEN)           = atol_temp ! temperature
```
maxpkatz: Why does SEDEN hold temperature?

PR #200, opened 2019-10-29
Helmholtz refactor

maxpkatz:
This PR splits Helmholtz up from one monolithic function into separate functions that add the radiation, ion, electron-positron, and Coulomb correction terms.

It also applies some optimizations to the electron-positron contribution involving the table read, including reusing some arrays to save on flops, and combining some tables to enhance coalesced memory accesses when reading the tabular data.


EOS/helmholtz/actual_eos.F90
```diff
@@ -1,1675 +1,1454 @@
...
-           deraddz = 0.0d0
-#endif
+    ! Reuse subexpressions that would go into computing h3.
+    wdt(1:4)   = sid(1) * sit(1:4)
+    wdt(5:8)   = sid(2) * sit(1:4)
+    wdt(9:12)  = sid(3) * sit(1:4)
+    wdt(13:16) = sid(4) * sit(1:4)
+
+    !..look in the pressure derivative only once
+    fi([ 1,  2,  5,  6]) = dpdf(1:4, iat,   jat  )
```
zingale: I don't understand this, shouldn't all of the indices that are at iat, jat be grouped here?  That would be [1, 5, 9, 13]

EOS/helmholtz/actual_eos.F90
```diff
@@ -1,1675 +1,1454 @@
...
+    if (input .eq. eos_input_rt) converged = .true.
 
-        double precision :: p_temp, e_temp
+    ! Iterate until converged.
 
-        double precision :: smallt, smalld
+    do iter = 1, max_newton
 
-        !$gpu
+       call apply_radiation(state)
```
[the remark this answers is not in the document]
maxpkatz: There was a comment inside apply_radiation and now I've added one above the call to it.

EOS/helmholtz/actual_eos.F90
maxpkatz: I chose to order `fi` not by the intuitive ordering you'd expect from reading in the points in the tabular data, but rather by the order in which they're used in the subsequent expression (`fi * wdt`, which replaces what `h3` used to do). If I reordered the entries in `fi` then I'd have to reorder `wdt` in a less clean way, so either way there's going to be weird indexing.

I'm not wedded to this specific way of evaluating it, and actually one of the things on my to-do list is to see if I can come up with anything better, but I figured we would get this PR done first before diving into that.

PR #201, opened 2019-11-02
Sync up VODE90 with VODE/BS retry strategy

maxpkatz:


integration/Make.package
```diff
@@ -12,7 +12,8 @@ else ifeq ($(INTEGRATOR_DIR),BS)
 else ifeq ($(INTEGRATOR_DIR),VBDF)
   INTEGRATOR_NUM := 2
 else ifeq ($(INTEGRATOR_DIR),VODE90)
-  INTEGRATOR_NUM := 3
+  # VODE90 is a drop-in replacement for VODE, treat it the same way.
+  INTEGRATOR_NUM := 0
```
zingale: I think we have some places in Castro where we check the value of this in if-defs, so changing 3 to 0 might break code.

maxpkatz: Renumbered to 3

zingale:
can you add a description to the CHANGES?

zingale:
also, there is a merge conflict -- is the resolution to just remove the old code?

PR #210, opened 2019-11-09
Modify the Jacobian contributions from the neutrino cooling

maxpkatz:
The neutrino cooling term returns d(S_nu)/dT, d(S_nu)/d(Abar), and d(S_nu)/d(Zbar). Since we're interested in dE/dY_j, we need to use the chain rule to obtain -dE/dY_j = d(Abar)/dY_j * d(S_nu)/d(Abar) + d(Zbar)/dY_j * d(S_nu)/d(Zbar). The latter term is correctly computed by all networks, since an analytic calculation reveals that d(Zbar)/dY_j is (Z_j - Zbar) * Abar. For the former term, it can either be -Abar^2, or (A_j - Abar) * Abar, depending on whether you enforce that sum(X) = 1 prior to taking the derivative.

Out of Frank Timmes' public networks as of this date, the 7 isotope and 34 isotope networks do it the way recommended by this PR, but the 13, 19 and 21 isotope networks do it the other way (the way that Microphysics does it, prior to this PR).

### DIFF
--- a/networks/ECSN/actual_rhs.F90
+++ b/networks/ECSN/actual_rhs.F90
@@ -426,7 +426,7 @@ contains
     call sneut5(state % T, state % rho, state % abar, state % zbar, sneut, dsneutdt, dsneutdd, snuda, snudz)
 
     do j = 1, nspec
-       b1 = ((aion(j) - state % abar) * state % abar * snuda + (zion(j) - state % zbar) * state % abar * snudz)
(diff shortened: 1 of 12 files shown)


maxpkatz:
We are still unclear which version is more correct. However numerically it seems to be better:

http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2019-11-10-005/index.html

PR #217, opened 2019-11-20
Add enthalpy evolution mode for simplified SDC

dwillcox:
To support the MAESTROeX style of simplified SDC, this PR adds an enthalpy evolution mode for the BS and VODE90 integrators with simplified SDC enabled.

The enthalpy evolution mode integrates an ODE system with rho*X and rho*h, the species partial densities and enthalpy.

This mode is turned on when `SIMPLIFIED_SDC` and `SDC_EVOLVE_ENTHALPY` are defined.

The Castro style energy evolution mode is enabled when `SIMPLIFIED_SDC` and `SDC_EVOLVE_ENERGY` are defined.


integration/BS/bs_jac.F90
```diff
@@ -33,7 +33,7 @@ subroutine jac(bs)
 
     if (jacobian == 1) then
 
-       call network_jac(bs % burn_s)
+       call network_jac(bs % burn_s, bs % upar(irp_t0))
```
zingale: so the network's Jac and rhs now take the starting time of the integration as an argument, not the current time?  That seems a little odd.

zingale: oh... I see, this is a wrapper around actual_jac.  That's not as bad.

PR #219, opened 2019-11-21
fix an index bug in the Jacobian temperature element

zingale:
The temperature element was being accessed incorrectly and we were overwriting with the nuclear energy generation


maxpkatz:
I believe the point of that code is that de/dX_k = c * dT/dX_k since e = c * T (where c is either c_V or c_P).

zingale:
this was not an issue

PR #220, opened 2019-11-22
remove the jac and ydot from burn_t

zingale:
This removes `jac` and `ydot` from the `burn_t` to save on memory and copying.


zingale:
this gives a ~ 5% speed up on the GPU for the 3-d flame_wave problem

networks/triple_alpha_plus_cago/actual_rhs.F90
```diff
@@ -117,7 +121,7 @@ subroutine actual_jac(state)
 
     ! initialize
     do j = 1, neqs
```
maxpkatz: It's possibly outside the scope of this PR, but since we're now dimensioning the Jacobian to have a number of columns that is not necessarily `neqs`, this loop should be from 1 to `njrows`.

zingale: right.  I think that we address this when we convert each network over to sparse.

integration/utils/jacobian_sparsity.F90
```diff
@@ -140,13 +140,19 @@ end subroutine get_csr_jac_entry
...
+
+    use burn_type_module, only : neqs
 
     !$acc routine seq
 
     implicit none
 
-    type (burn_t), intent(inout) :: state
+#ifdef REACT_SPARSE_JACOBIAN
+    real(rt), intent(inout) :: jac(NETWORK_SPARSE_JAC_NNZ)
```
maxpkatz: We can do this now without an ifdef, right?

PR #225, opened 2019-12-06
use a kind to make precision flexible

zingale:


zingale:
we're going to do this in stages

PR #227, opened 2019-12-07
make precision more flexible

zingale:
This gets rid of all of the double precision stuff in favor of the rt type.  The conversion was done with the dpconvert.py script from Castro.


zingale:
all tests pass

zingale:
going to rerun the script off of development and create a new PR

Issue #147, opened 2018-08-26
original SDC implementation Jacobian is not correct


zingale:
The analytic Jacobian for the original SDC (enabled by setting the `SDC` preprocessor) is not correct.  The Jacobian elements should be derivatives with respect to the conserved variables.  The proper way to do this is to construct a second transformation Jacobian, d(rho, rho X, rho E)/d(rho, T, X) and use this to transform the derivatives that come from the burner.  This is actually quite easy, and is implemented in the new SDC implementation in the `sdc` branch of Castro.

With the analytic Jacobian enabled (`jacobian=1`), this original SDC method surprisingly runs, but very slowly.

zingale:
At the moment, our state vector that the ODE integration sees is (rho X, rho E, rho e).  We get the Jacobian from the net in terms of rho, T, X.  I think in order to do the transformation properly, we should integrate only one of (rho E) or (rho e), and add rho to the ODE integration vector.  Then we can do the same transformation as we are doing in the new SDC stuff.

zingale:
It looks like the function that needs to be modified is `jac_to_vode` in `vode_type_simplified_sdc.F90`

PR #228, opened 2019-12-09
fix the Jacobian for the simplified-SDC system

zingale:
For simplified SDC we need to convert from the reaction system Jacobian (in terms of X, T, e) to the system for the conserved variables.  This does that transformation.


zingale:
this will close #147 

still need to do the MAESTROeX / enthalpy part

zingale:
This now works for both the energy and enthalpy-based SDC formulations.

zingale:
The energy based Jacobian derivation is here:

[simplified_sdc_derivations.pdf](https://github.com/starkiller-astro/Microphysics/files/3969724/simplified_sdc_derivations.pdf)

The enthalpy based Jacobian derivation is here:

[maestroex_simplified_sdc_derivations.pdf](https://github.com/starkiller-astro/Microphysics/files/3969730/maestroex_simplified_sdc_derivations.pdf)

zingale:
For the `test_sdc` unit test, with these changes, when we run with the numerical Jacobian, we get:
```
Run time = 19.73930645
min number of rhs calls: 21
avg number of rhs calls: 445
max number of rhs calls: 6883
```
and with the analytic Jacobian we get:
```
Run time = 17.52168508
min number of rhs calls: 6
avg number of rhs calls: 358
max number of rhs calls: 5737
```

The speed-up is the same when considering the enthalpy system.  On `development`, with the analytic Jacobian, we get:
```
Run time = 20.20451861
min number of rhs calls: 6
avg number of rhs calls: 450
max number of rhs calls: 12023
```

so this change is definitely a win.

zingale:
I think that this is ready to be reviewed.  Here's an updated derivation of the system used for energy.

[simplified_sdc_derivations.pdf](https://github.com/starkiller-astro/Microphysics/files/4554839/simplified_sdc_derivations.pdf)

zingale:
Castro detonation tests, running for 250 steps.

* numerical Jacobian:

  Run time without initialization = 97.3357413

* analytic Jacobian with Microphysics on ``development``:

  Run time without initialization = 2439.936833

* analytic Jacobian with Microphysics on ``sdc_jacobian`` before my most recent changes
  (``9d097e745ab5c00620546f7e0d154df34b1a98f3``):

  dies with burn failures right out of the gate in lots of different zones

* analytic Jacobian with Microphysics on ``sdc_jacobian`` after my
  changes:

  Run time without initialization = 66.74610559

PR #236, opened 2019-12-16
run the dpconvert.py script to fix precision issues

zingale:


zingale:
that was an older version of the script

PR #239, opened 2019-12-17
Convert all floats to _rt

harpolea:
This converts all values with decimal points in them to be of the form #.#e#_rt


EOS/ztwd/actual_eos.F90
```diff
@@ -4,8 +4,8 @@
 ! 
 ! where rho = B x**3 and the constants are given by:
 !
-! A = pi m_e**4 c**5 / (3 h**3) = 6.0 x 10^22 dyne cm**-2
-! B = 8 pi m_e**3 c**3 mu_e m_p  / (3 h**3) = 9.8 x 10^5 mu_e g cm**-3
+! A = pi m_e**4 c**5 / (3 h**3) = 6.0_rt x 10^22 dyne cm**-2
+! B = 8 pi m_e**3 c**3 mu_e m_p  / (3 h**3) = 9.8_rt x 10^5 mu_e g cm**-3
```
zingale: don't need to change comments

integration/BS/bs_type.F90
```diff
@@ -6,7 +6,7 @@ module bs_type_module
 
   implicit none
 
-  ! BS parameters -- see the discussion in 16.4
+  ! BS parameters -- see the discussion in 16.4_rt
   integer, parameter :: KMAXX = 7
```
zingale: comment

integration/BS/stiff_ode.F90
```diff
@@ -633,7 +633,7 @@ subroutine single_step_bs(bs, eps, yscal, ierr)
     bs % first = .false.
 
     ! optimal convergence properties
-    work_min = 1.e35
+    work_min = 1.e35_rt
```
zingale: maybe we should use `huge()` here?

networks/ECSN/reaclib_rates.F90
```diff
@@ -163,9 +163,9 @@ subroutine reaclib_evaluate(pstate, temp, iwhich, reactvec)
           dlnirate_dt = dlnirate_dt + &
                T9_exp * ctemp_rate(j, istart+i) * T9**(T9_exp-1.0e0_rt)
        end do
-       ! If the rate will be in the approx. interval [0.0, 1.0E-100], replace by 0.0
+       ! If the rate will be in the approx. interval [0.0_rt, 1.0e-100_rt], replace by 0.0_rt
        ! This avoids issues with passing very large negative values to EXP
-       ! and getting results between 0.0 and 1.0E-308, the limit for IEEE 754.
+       ! and getting results between 0.0_rt and 1.0e-308_rt, the limit for IEEE 754.
        ! And avoids SIGFPE in CVODE due to tiny rates.
```
zingale: we might want to handle these limits differently for different precision

Issue #242, opened 2020-01-29
create C++ EOSes


zingale:
To move Castro's hydro off of Fortran, we need a C++ EOS.  There are a few approaches we can consider:

  1. Make a separate implementation of each EOS in C++, so we would have a separate complete Fortran and C++ EOS routines.

  2. Create a C++ wrapper to the existing EOSes.  It might be hard to keep the current Fortran derived type, so we might need an intermediate interface that accepts a C++ struct and unpacks it into the Fortran type.

3. Get rid of the Fortran implementations completely and rewrite all the EOSes in C++.  Then, if needed, we could create Fortran interfaces to call the C++

maxpkatz:
I prefer option 1 since we cannot move everything to C++ at one time. I do not believe we spend that much time on EOS maintenance, and it sounds less tricky than getting Fortran to C++ (or vice versa) interfaces correct.

maxpkatz:
The interface for C should mirror Fortran; the eos_t is a C struct, and we still do eos(input, state).

zingale:
the framework for this has been merged.  Now it's just adding semi-colons ;)

PR #244, opened 2020-01-31
add nuclei into to the C++ header

zingale:
This updates the actual_network.H file to include aion and zion.  A .net file is present in each network now that creates the nuclei information for both Fortran and C++.  This is done by running the top-level update_headers.py script.  That writes actual_network.H and a new module, network_parameters.F90.  The actual_network.F90 has been rewritten to use network_parameters.F90.

All unit tests pass with this change.


zingale:
This PR is needed for the C++ EOSes that are part of #242

maxpkatz:
I think we need to rename the auto-generated C++ file to network_properties.H, to match the Fortran. Then actual_network.H would #include it. This will allow us to later fill in more data in actual_network.H when we eventually do a full C++ implementation of the networks.

zingale:
I've made this change.

PR #253, opened 2020-02-16
autogenerate network_properties.{F90,H}

zingale:
This adds a rule to networks/Make.package that generates the files network_properties.F90 and network_properties.H at build time.  You can set the location to output to by defining NETWORK_HEADERS_DIR.  Note: the pynucastro generated networks will need to be updated to output the .net file.


zingale:
all tests pass
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-02-16-004/index.html

PR #256, opened 2020-02-16
more VODE opts

zingale:
Pass the dvode_t directly into the conversion routines to save on memory.


zingale:
all tests pass:
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-03-11-008/index.html

dwillcox:
I bet merging development into vode_opts2 will fix these conflicts, these are probably from merging #279

PR #260, opened 2020-02-17
keep only a single specific heat

zingale:
We don't need both cv and cp, just whichever one is being used for T evolution.


zingale:
this works for vode, but I am not sure this is worth it in the end.  We still need to sprinkle the logic about do_constant_volume burn throughout the code when passing back and forth with the EOS.  I didn't think some of those we needed, but it turns out they are.

zingale:
this is more complicated than I like and probably not worth it.

PR #261, opened 2020-02-20
restructure makefiles

zingale:
This reorganizes the makefiles to separate out the bits that include all of the various Microphysics modules.  This will make it easier for Castro to simply include `Make.Microphysics_extern` instead of duplicating it all on its own.

### DIFF
--- a/Make.Microphysics_extern
+++ b/Make.Microphysics_extern
@@ -1,3 +1,5 @@
+# This is the main include makefile for applications that want to use Microphysics
+# You should set NETWORK_OUTPUT_PATH before including this file
 
 EOS_PATH := $(MICROPHYSICS_HOME)/EOS/$(strip $(EOS_DIR))
 NETWORK_PATH := $(MICROPHYSICS_HOME)/networks/$(strip $(NETWORK_DIR))
--- a/Make.Microphysics_extern
+++ b/Make.Microphysics_extern
@@ -1,0 +1,71 @@
+
+EOS_PATH := $(MICROPHYSICS_HOME)/EOS/$(strip $(EOS_DIR))
+NETWORK_PATH := $(MICROPHYSICS_HOME)/networks/$(strip $(NETWORK_DIR))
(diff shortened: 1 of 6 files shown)


zingale:
everything seems to build

integration/Make.package
```diff
@@ -19,6 +19,13 @@ endif
 
 DEFINES += -DINTEGRATOR=$(INTEGRATOR_NUM)
 
+ifeq ($(USE_CUDA), TRUE)
+  ifneq ($(INTEGRATOR_DIR), VODE90)
```
maxpkatz: This check should really be on `VODE`, not `VODE90`.

networks/Make.package
```diff
@@ -7,38 +7,35 @@ include $(MICROPHYSICS_UTIL_PATH)/Make.package
 
 ifeq ($(USE_REACT), TRUE)
 
-ifeq ($(INTEGRATOR_DIR), VODE90)
-  override INTEGRATOR_DIR = VODE
-  $(warn VODE90 has been renamed VODE)
-endif
+  ifeq ($(INTEGRATOR_DIR), VODE90)
```
maxpkatz: Should this code now be in integration/Make.package?

Make.Microphysics_extern
```diff
@@ -0,0 +1,89 @@
+# This is the main include makefile for applications that want to use Microphysics
+# You should set NETWORK_OUTPUT_PATH before including this file
```
maxpkatz: Can we do `NETWORK_OUTPUT_PATH ?= $(CURDIR)` in case they didn't set it?

zingale: No, it turns out that breaks things.  They way it was done previously was not ever being used.  I think we can do `NETWORK_OUTPUT_PATH ?= .` though.

networks/Make.package
zingale: probably.  I tried not to dive too deeply into the module make stubs yet, so I would prefer to clean them on a second pass.

integration/Make.package
zingale: yes.  I thought I fixed that.

PR #267, opened 2020-02-24
License

zingale:
add licensing info where it is known


adam-m-jcbs:
Looks way better than what I see many people do.  Glad we're paying attention to this, even if too few others are.

dwillcox:
Thanks for the license and for tracking down all the license information for the code we distribute!

PR #269, opened 2020-02-27
Convert rad_power_law to C++

maxpkatz:


EOS/rad_power_law/actual_eos.H
```diff
@@ -0,0 +1,91 @@
...
+
+        state.cv = eos_const_c_v * std::pow(state.rho, eos_c_v_exp_m) * std::pow(state.T, -eos_c_v_exp_n);
+        state.e  = eos_const_c_v * std::pow(state.rho, eos_c_v_exp_m) * std::pow(state.T, 1 - eos_c_v_exp_n) / (1 - eos_c_v_exp_n);
+
+        break;
+
+    case eos_input_re:
+
+        state.T  = std::pow((1 - eos_c_v_exp_n) * state.e * std::pow(state.rho, -eos_c_v_exp_m) / eos_const_c_v, 1 / (1 - eos_c_v_exp_n));
+        state.cv = eos_const_c_v * std::pow(state.rho, eos_c_v_exp_m) * std::pow(state.T, -eos_c_v_exp_n);
```
zingale: maybe do 1.0_rt here to be safer?

maxpkatz: The exponent is an int so I wanted to keep it that way from the Fortran in case std::pow is more efficient.

zingale: but your exponent is `1 / (1 - eos_c_v_exp_n)`, so integer division might not give you what you want here.

maxpkatz: Yes, that's true. However I was wrong, we are saving those exponents as Real, not as int, so it was actually fine. Nevertheless, I did make that change since it does not hurt to be safe.

PR #271, opened 2020-03-05
Place the Helmholtz globals in a namespace

zingale:
this makes it less likely to get shadow warnings


EOS/helmholtz/actual_eos_data.H
```diff
@@ -4,60 +4,65 @@
...
 #include <AMReX_REAL.H>
 
-extern AMREX_GPU_MANAGED bool do_coulomb;
-extern AMREX_GPU_MANAGED bool input_is_constant;
+namespace helmholtz
+{
 
-// for the tables
+    extern AMREX_GPU_MANAGED bool do_coulomb;
+    extern AMREX_GPU_MANAGED bool input_is_constant;
```
zingale: do we need separate copies of these?  We can use the parameters from extern_parameters.H directly, can't we?

maxpkatz: That is right. A long time ago in the Fortran I made a copy in a weak attempt at proper OOP so that the EOS would own whater its internal data was. But it really does not matter much to me.

PR #275, opened 2020-03-11
remove some unused code paths and go tos

zingale:
all tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-03-11-002/index.html


integration/VODE/cuVODE/source/cuvode_dvhin.F90
```diff
@@ -118,63 +125,66 @@ subroutine dvhin(vstate, rwork, H0, NITER, IER)
...
-    ENDIF
-    ITER = ITER + 1
-    ! -----------------------------------------------------------------------
-    !  Test the stopping conditions.
-    !  Stop if the new and previous h values differ by a factor of .lt. 2.
-    !  Stop if four iterations have been done.  Also, stop with previous h
-    !  if HNEW/HG .gt. 2 after first iteration, as this probably means that
-    !  the second derivative value is bad because of cancellation error.
-    ! -----------------------------------------------------------------------
-    IF (ITER .GE. 4) GO TO 80
```
maxpkatz: We can still keep this as `if (ITER .ge. 4) exit`, and this logic is a little complicated so I can't say for sure what your `while` condition captures this properly.

Issue #277, opened 2020-03-12
get rid of nspec_evolve < nspec


zingale:
We originally had allowed for unevolved species for ignition_simple which was written back for the original MAESTRO.  There we had C12, O16, Mg24, but only needed to react the C12.  But we never use this for networks with more than 4 species, and the smaller nets are fast enough even if we evolve everything.

So we should drop this feature.  It's a pain to support and it doesn't work with SDC.  

If we remove this, we can also get rid of the separate `xn(:)` in the `burn_t`, which will free up more memory, since we need to copy it from VODE's `y` into the `burn_t`.

PR #279, opened 2020-03-12
get rid of nspec_evolve

zingale:
This addresses #277 -- we remove the ability to have nspec_evolve < nspec.  We only ever used it 
for the smallest nets, and it didn't work with simplified SDC.  

Note: for the nets that did have nspec_evolve < nspec, we expect tolerance-level changes to the answers because the ODE integrator is now working with a different system.

http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-03-12-002/index.html

### DIFF
--- a/integration/BS/bs_integrator.F90
+++ b/integration/BS/bs_integrator.F90
@@ -36,7 +36,6 @@ contains
 
     use bs_rpar_indices
     use extern_probin_module, only: burner_verbose, burning_mode, burning_mode_factor, dT_crit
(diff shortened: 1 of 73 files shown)


zingale:
Note: I didn't update the C++ cvode stuff.

zingale:
we'll either need to modify Castro and Maestro to not refer to nspec_evolve before we merge this, or I can add a temporary `nspec_evolve = nspec` to the main `network.F90` to cover the transition.

PR #281, opened 2020-03-17
Vode goto 5

zingale:
this eliminates more go tos:

http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-03-17-002/index.html


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-03-18-002/index.html

PR #282, opened 2020-03-18
Vode goto 6

zingale:
more go to elimination


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-03-18-004/index.html

PR #284, opened 2020-03-18
Remove gotos in linpack_module

maxpkatz:


integration/VODE/cuVODE/source/linpack_module.F90
```diff
@@ -74,63 +74,61 @@ subroutine dgesl(a, ipvt, b)
...
-    ! 
+
     real(rt)         t
     integer k,kb,l,nm1
+
     !$gpu
-    ! 
+
     nm1 = n - 1
-    if (job .ne. 0) goto 50
```
zingale: do we need the job !=- 0 logic?

zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-03-18-005/index.html

Issue #266, opened 2020-02-24
unify test_eos and test_eos_C


zingale:
The C++ and Fortran EOS unit tests should be merged _and_ the test should call both and compare them to ensure they agree

PR #292, opened 2020-03-26
merge C++ and Fortran unit tests

zingale:
The merges the test_eos unit tests


zingale:
addresses #266

zingale:
okay, this works, and the benchmarks for the Fortran and C++ are the same as before, but they are just named differently so the suite will complain.  But once this is merged, it'll all be happy.

unit_test/test_eos/input_eos
```diff
@@ -1,4 +1,4 @@
 n_cell = 16
 max_grid_size = 32
 
-amr.probin_file = probin
+amr.probin = probin
```
maxpkatz, line 4: `amr.probin_file` is the correctname of the parameter, what's up here?

zingale: you're right -- that was a typo, but since the default is "probin", it was still getting the right probin file

maxpkatz:
> 
> 
> okay, this works, and the benchmarks for the Fortran and C++ are the same as before, but they are just named differently so the suite will complain. But once this is merged, it'll all be happy.

We could keep the plotfile names the same, couldn't we?

zingale:
yeah, I could keep them the same -- I was trying to follow the convention used in test_react and the other unit tests.  But I'll make them have the same name as before for now and then we can sync things up later.

Issue #289, opened 2020-03-25
eliminate screenz in favor of screen5


zingale:
only a few nets use screenz.  it has not been as optimized as screen5, so we should switch those nets over to screen5 and remove screenz.

PR #293, opened 2020-03-27
eliminate screenz

zingale:
Only 3 networks were using screenz (triple_alpha_plus_cago, ignition_simple, ignition_chamulak).
We swap them over to screen5.


zingale:
This addresses #289

zingale:
diffs are quite small:

http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-03-26-001/index.html

PR #302, opened 2020-04-14
fix O16+O16 screening setup

zingale:
this was caught by the C++ assert


dwillcox:
Agreed! O16 + O16 is the corresponding rate on line 647, (6th call to screen5).

Thanks!

PR #303, opened 2020-04-14
Move iso7 to C++

dwillcox:
This PR moves the iso7 network to C++ as well as other parts of Microphysics the network depends on.

This includes a C++ version of `sneut5` as well as the `esumN` code generator script.

A test is added in `unit_test/test_rhs` that calls the RHS and Jacobian for a network and saves the output into a plotfile using either the C++ or Fortran network to make comparisons easy.

Attached is the fcompare for the RHS and Jacobian for iso7 using the same thermodynamic ranges as inputs as for the `test_react` unit test.

[fc.txt](https://github.com/starkiller-astro/Microphysics/files/4473040/fc.txt)

This PR also rearranges the screening API a bit to require zion and aion for all screened nuclei for convenience.

Right now the interfaces like `burn_type` and `eos_type` as well as `aion`, `aion_inv`, and `zion` still use 0-based indexing. The C++ iso7 liberally uses `amrex::Array1D` with 1-based indexing to make comparisons & porting from Fortran much easier.

In the future we can standardize this, but that's probably best saved for another PR.


networks/iso7/actual_network.H
```diff
@@ -0,0 +1,73 @@
...
+
+        constexpr amrex::Real n_A = 6.0221417930e23_rt;
+
+        // conversion factor for nuclear energy generation rate
+        constexpr amrex::Real enuc_conv2 = -n_A * c_light * c_light;
+    }
+}
+
+namespace Species {
+    enum NetworkSpecies {He4 = 1, C12, O16, Ne20, Mg24, Si28, Ni56, NumberSpecies = Ni56};
```
maxpkatz, line 36: @zingale can we have the network_properties.H auto-generate an enum like this?

zingale: yes.  I'll do that in a branch right now.  We want it 1-based?

maxpkatz: Yes, 1-based for now, thank you.

zingale: do we really want NumberSpecies in the enum too?

networks/iso7/actual_rhs.H
```diff
@@ -0,0 +1,947 @@
...
+                        Array2D<Real, 1, neqs, 1, neqs>& jac)
+{
+
+    // this routine sets up the dense iso7 jacobian for the isotopes
+
+    using namespace Species;
+    using namespace Rates;
+
+    {
+        Array1D<Real, 1, 8> b;
```
maxpkatz, line 455: In the Fortran what we were doing here is reusing the same array and relying on the fact that esumN ever uses more than N elements of the array. I don't have any problem with this way of writing it, as long as the compiler figures out that it can reuse the registers under the hood.

networks/iso7/actual_network.H
maxpkatz: Well it can't hurt to throw NumSpec in the enum, but we don't need it.

zingale: see https://github.com/starkiller-astro/Microphysics/pull/304

unit_test/test_rhs/variables.H
```diff
@@ -0,0 +1,21 @@
...
+enum plot_t
+{
+    p_irho = 0,
+    p_itemp = 1,
+    p_ispec = 2,
+    p_ispec_old = p_ispec + NumSpec,
+    p_itemp_dot = p_ispec_old + NumSpec,
+    p_ienuc_dot = p_itemp_dot + 1,
+    p_ijac = p_ienuc_dot + 1,
+    n_plot_comps
```
maxpkatz, line 18: Don't think it matters, but n_plot_comps should be offset by neq**2

networks/iso7/actual_network.H
dwillcox: Thanks! I've removed the Species namespace/enum after merging #304 into development and then development into this PR branch.

unit_test/test_rhs/variables.H
dwillcox: Fixed, thanks!

networks/iso7/actual_rhs.H
dwillcox: yes, hopefully so with the {} scoping

it's easy enough to make the esums take arbitrary sized Array1Ds if we need

PR #305, opened 2020-04-16
Move aprox13 to C++

dwillcox:
This PR ports aprox13 to C++ using the same strategy as for iso7 in #303 

Here are the fcompare results between C++ and Fortran for the `test_rhs` unit test:

[fc_aprox13_rhs.txt](https://github.com/starkiller-astro/Microphysics/files/4484423/fc_aprox13_rhs.txt)

Here are the fcompare results between C++ and Fortran for the `test_react` unit test:

[fc_aprox13_react.txt](https://github.com/starkiller-astro/Microphysics/files/4484426/fc_aprox13_react.txt)

This PR also adds input files for `test_rhs` for aprox13 based on test_react inputs, except I've added `call_eos_in_rhs=T`.

To avoid completely refactoring the rhs in aprox13, I also enabled passing arbitrary size `Array1D` arrays to all esum functions by making esums templated functions.

I've also run the `test_rhs` and `test_react` unit tests in debug mode and successfully got the exact same results as without debug mode.

### DIFF
--- a/networks/aprox13/actual_network.H
+++ b/networks/aprox13/actual_network.H
@@ -16,7 +16,7 @@ namespace C
 {
     namespace Legacy
     {
-        // These are the values of the constants used in the original iso7
+        // These are the values of the constants used in the original aprox13
         constexpr amrex::Real m_n = 1.67492721184e-24_rt;
         constexpr amrex::Real m_p = 1.67262163783e-24_rt;
         constexpr amrex::Real m_e = 9.1093821545e-28_rt;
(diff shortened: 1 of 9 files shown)


networks/aprox13/actual_network.H
```diff
@@ -0,0 +1,120 @@
...
+
+using namespace amrex;
+
+void actual_network_init();
+
+namespace C
+{
+    namespace Legacy
+    {
+        // These are the values of the constants used in the original iso7
```
maxpkatz, line 19: Comment should be fixed (iso7 -> aprox13)

dwillcox: Fixed, thanks!

PR #307, opened 2020-04-16
Use sparse storage & operations in C++ VODE for aprox13

dwillcox:
This PR adds a sparse linear system storage and arithmetic to C++ VODE for aprox13.

The basic idea is to create sparse and dense matrix data structures for the Jacobian so the rhs and integrator can interact with it through provided functions agnostic to the underlying storage.

- Adds a sparse jacobian struct for aprox13 that handles 2D->1D indexing for access and arithmetic at compile time.

- Adds a dense jacobian struct (`MathArray2D`) that includes the equivalent arithmetic operations we need to supply for a consistent interface between sparse & dense jacobian data structures.

- Modifies rhs, temperature rhs, and integrator to be agnostic to the storage type

- Modifies `test_rhs` to use the sparse or dense jacobians to call `actual_jac` depending on `USE_NETWORK_SOLVER`.

Here are the fcompare results vs Fortran for `test_rhs` using the sparse Jacobian: 
[fcsparse.txt](https://github.com/starkiller-astro/Microphysics/files/4489117/fcsparse.txt)

Here are the fcompare results vs Fortran for `test_react` using the sparse Jacobian: 
[fc_sparse_aprox13.txt](https://github.com/starkiller-astro/Microphysics/files/4489145/fc_sparse_aprox13.txt)

This is ready for review, pending the merge of #305


dwillcox:
Just fixed a build issue for iso7.

Right now iso7 does not support `USE_NETWORK_SOLVER=TRUE` because it doesn't define a sparse jacobian, saving that for future work.

Here's c++ vs fortran for the iso7 rhs: 
[fc_iso7_rhs.txt](https://github.com/starkiller-astro/Microphysics/files/4491075/fc_iso7_rhs.txt)

dwillcox:
Just rebased onto development and switched PR from draft to ready for review.

PR #328, opened 2020-06-23
C++ and Fortran Comparison

shardi2:
Here are my changes to the C++ and Fortran burn_cell code.


unit_test/burn_cell/burn_cell.F90
```diff
@@ -126,8 +126,21 @@ subroutine burn_cell() bind(C, name="burn_cell")
   call eos(eos_input_rt, eos_state_in)
   call eos_to_burn(eos_state_in, burn_state_in)
 
-  dt = tmax
-  call actual_burner(burn_state_in, burn_state_out, dt, time)
+  open(unit=1, file="state_over_time_F90.txt")
+  10 format(1x, f10.9)
+  20 format(1x, f4.0/)
+
+  dt = nsteps
```
zingale, line 133: I think that since we take `tmax` already, we should use `nsteps` to compute `dt` as `dt = tmax / nsteps`, then we pass that `dt` into the burn routine.  This will allow us to control the timstep instead of hardcoding it to `1`

unit_test/burn_cell/burn_cell.F90
```diff
@@ -126,8 +126,21 @@ subroutine burn_cell() bind(C, name="burn_cell")
...
+     n = 1.0
+     call actual_burner(burn_state_in, burn_state_out, n, time)
+     call copy_burn_t(burn_state_in, burn_state_out)
+     yloop: do y = 1, nspec
+        write(1, 10, advance="no") burn_state_out % xn(y)
+     end do yloop
+     write(1, 20)
+  end do xloop
+  close(1)
   energy = energy + burn_state_out % e
```
zingale, line 145: I am not 100% whether this energy update should be inside the loop, so we increment the energy during each of the steps

zingale:
can you add your plotting script to this as well?

zingale:
I think this looks good now.

Can you add temperature to the state_over_time.txt output (as another column)?

Also, can you get rid of the blank line between the different rows in the output?

Issue #327, opened 2020-06-22
NSE table


zingale:
The `NSE` branch has the NSE tables from Stan.  

from Stan:

`nse19.tbl` : a table with the standard aprox19 isotopes

`usetable.f` : a driver for nse19.tbl

`castro-nse-table` : a version without all of the isotopes

`usetable_small.f` : a driver for castro-nse-table

from Stan:

Probably all the code needs for hydrodynamics and energy generation is
abar and Ye (given rho and T) so to follow explosions you can use the
smaller table. It also gives crude measures of nucleosyntheis in
n+p+helium, si-ca, fe group. Probably good enough for NSE

If you are following bulk nucleosynthesis you can use the larger table
which gives 125 isotopes packaged into 19 (CASTRO used to use the 19
iso network of KEPLER, maybe it no longer does?). These abundances can
be interpolated by a simple extension of the usetable.f routine to
more variables

abund(18) is neutrons
abund(19) is protons
1 is H1  is 0
2 is mostly 3He
3                4He
4               12C
5                14N
6                16O
7                20Ne
8                24Mg
9                28Si
10              32S
11              36Ar
12              40Ca
13              44Ti
14              48Cr
15              52Fe
16              all oher iron group except 56Ni
17              56Ni

The Ye of those 19 isotopes may not agree exactly with the Ye of the
table and the latter should be used in the hydro and EOS. In fact
since the A of abund 16 is not given, Ye is indeterminate.

Carrying all these isotopes merely adds to the memory so long as you
are in nse. But if you are following a continuum of burning from He (or
H) to NSE you may need them.  In fact, I think all burning stages can
be handled with tables e.g. for carbon burning makes tables of
composition for carbon burned at constant T and rho and interpolate in
X12, T and rho, though some thought must e given to the initial carbon
abundance.

zingale:
Stan also sent me the full network table (it is > 150 MB), but I don't think we want to use that.

So it seems like we should use the nse19.tbl table together with aprox19, and this then will give us the species in each zone (approximately) and allow use to continue to advect stuff.

zingale:
I think what we actually want to do is couple this to our aprox19 network.  So that it automatically calls the table and updates things if the thermodynamic conditions warrant -- all through the normal interface.

We will also need to have the EOS get ye and abar, but perhaps we can, as a first pass, hijack the EOS composition routine from this network to override Abar and Ye via another table call (assuming table calls are cheap).

[...]

PR #332, opened 2020-06-27
add a hybrid aprox19 + NSE network

zingale:
This will use an NSE table at high densities and temperature to provide the energy generation and composition variables.


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-07-15-002/index.html

zingale:
this implements #327

networks/aprox19_nse/actual_burner.F90
```diff
@@ -0,0 +1,88 @@
...
+
+       ! under-relaxation / inertia (see Ma et el. 2013)
+       deltaq = 0.3_rt * deltaq
+
+       state_out % aux(ibea) = state_in % aux(ibea) + deltaq
+
+       ! convert the energy to erg / g
+       enuc = deltaq * mev2erg * avo
+
+       state_out % e = enuc - state_in % e
```
maxpkatz, line 67: I think this should be a plus?

networks/aprox19_nse/actual_burner.F90
```diff
@@ -0,0 +1,88 @@
...
+
+       state_out % aux(iye) = state_in % aux(iye) - dt * dyedt
```
maxpkatz, line 69: Just double checking that this is correct -- dye/dt is negative? Can we just reverse it in the nse_interp routine so that this is less confusing?

maxpkatz:
I suspect it would be simpler to just merge this into aprox19 rather than have duplicated networks. It looks like the changes relative to aprox19 are minimal. The changes to the integrator are minimal and easy to merge. And we could just add the NSE files. The only part that I think conflicts is that we add aux variables to the network. Could those aux variables be conditioned on whether we're using the NSE bits?

networks/aprox19_nse/actual_burner.F90
```diff
@@ -0,0 +1,88 @@
...
+
+    call in_nse(state_in, nse_check)
+
+    if (nse_check == 0) then
+
+       ! integer the aprox19 network
+
+       call integrator(state_in, state_out, dt, time)
+
+       ! update the auxillary state
```
maxpkatz, line 39: auxiliary

networks/aprox19_nse/actual_burner.F90
zingale: yes, you are right (good think it's 0 :)

networks/aprox19_nse/actual_burner.F90
zingale: it is negative.  According to Stan (as stated in the README.md):
What is tabulated is wrate which is the sum of all electron capture
and positron decay rates times the appropriate abundances minus a
similar rate for beta decay and positron capture

 wrate=rectot+rpdtot-redtot-rpctot

So if electron capure domiantes wrate is positive

The inference is that it is a positive term that is subtracted from Ye

zingale: I'll fix this in the interp routine -- I agree it is cleaner that way

networks/aprox19_nse/actual_burner.F90
zingale: fixed

zingale:
I agree it would be clearer to have this piggy back on aprox19.  I just need to understand how to make the aux stuff conditional -- at the moment, our scripts don't support that.

zingale:
`aprox19_nse` and `aprox19` are now merged and this still gives the same answer.  To enable NSE mode, you compile with `USE_NSE = TRUE`.  Together with the changes in #370, this will automatically add the needed AUX quantities to the network when using NSE.

This should be merged _after_ #370.

zingale:
all tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-08-07-002/index.html

networks/aprox19/nse.F90
```diff
@@ -0,0 +1,199 @@
+module nse_module
+
+  use amrex_fort_module, only : rt => amrex_real
+
+  implicit none
+
+  integer, parameter :: ntemp = 71
```
maxpkatz, line 7: Are these parameters used outside the module? If not can we declare them explicitly private for safety?

zingale: good call, I'll do that

zingale: I'll also pull nspec from network

maxpkatz:
Can you explain what AuxZero is?

networks/aprox19/nse.F90
zingale: updated

maxpkatz (approved):

PR #334, opened 2020-06-30
New C++ and Fortran Comparison

shardi2:
Redo of pull request


unit_test/burn_cell/_parameters
```diff
@@ -3,11 +3,13 @@ run_prefix    character  ""
 small_temp    real       1.e5
 small_dens    real       1.e5
 
-tmax          real       1.0d0
+tmax          real       1.e-2
 
-density       real       1.d4
+nsteps        real       1000
```
maxpkatz, line 8: You can use `integer` as a parameter instead of `real` for this quantity.

zingale:
thanks for your first contribution Samantha!

Issue #300, opened 2020-04-04
burning_mode = 0 and 1 can be removed


maxpkatz:
They are redundant with the `integrate_temperature` option, and that one is much easier to understand what it does.

maxpkatz:
Also, burning_mode = 2 (which retries a burn with temperature integration turned off if the self-heating burn fails) is not necessary now that Castro has retries, and burning_mode = 3 is not something we want to use in the long run anyway (it was an experiment to see if we want to use the burning limiters in the literature, and we do not), so we can just ditch the burning_mode parameter entirely.

zingale:
I'm confused by the title -- we don't want to remove burning_mode = 1, which is the default, but instead that is just going to be the only mode, right?

maxpkatz:
> 
> 
> I'm confused by the title -- we don't want to remove burning_mode = 1, which is the default, but instead that is just going to be the only mode, right?

Correct, I'm talking about removing the parameter, not the physics.

PR #340, opened 2020-07-11
remove burning mode -- we always assume we are burning_mode = 1

zingale:
this addresses #300 and greatly simplifies the integrators.


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-07-13-001/index.html

the EOS failures are because development is ahead of this change
the burning_mode 0 failure is expected, since we've taken that out

PR #342, opened 2020-07-15
eliminate some divides

zingale:


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-07-15-001/index.html

PR #343, opened 2020-07-16
unify the C++ and Fortran screening unit tests

zingale:
Previously we had test_screening and test_screening_C, but now the implementations are merged under test_screening.


zingale:
tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-07-17-003/index.html

unit_test/test_screening/main.cpp
```diff
@@ -98,24 +110,46 @@ void main_main ()
...
+      // get the variable names
+      for (int i=0; i<Ncomp; i++) {
+        char* cstring[name_len+1];
+        get_var_name(cstring, &i);
+        std::string name(*cstring);
+        varnames.push_back(name);
+      }
     }
 
+    std::cout << "Ncomp = " << Ncomp << std::endl;
```
maxpkatz, line 151: Debugging print to be removed?

zingale: done

PR #345, opened 2020-07-16
unify the conductivity unit tests (C++ and Fortran) into a single test

zingale:


zingale:
test here: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-07-17-006/index.html
looks like we never had a C++ test in the suite

PR #344, opened 2020-07-16
unify the C++ and Fortran test_aprox_rates unit tests

zingale:

PR #346, opened 2020-07-17
update the docs on the unit tests

zingale:


zingale:
This should wait for #343, #344, and #345

Issue #106, opened 2018-05-06
Too many failure modes for VODE?


maxpkatz:
Consider the following example of a failure from VODE:

```
 ERROR: integration failed in net                                                                                                                                                                            
 istate =  2                                                                                                                                                                                                 
 time =  4.71541112653262647E-3                                                                                                                                                                              
 dens =  33886998.007741921                                                                                                                                                                                  
 temp start =  5345296728.7418671                                                                                                                                                                            
 xn start =  4.28495683615790456E-3,  2.12569305413088262E-3,  0.46827461056448705,  3.93911881553560446E-3,  0.16770396099415991,  0.35367165973452869,  1.00000027059444347E-12                            
 temp current =  5391881235.7946024                                                                                                                                                                          
 xn current =  1.16034631797376116E-2,  1.19633015016508372E-3,  0.45320340626185762,  -2.50830974789474914E-2,  0.14005418455012586,  0.41902571333606137,  1.00000027059444326E-12                         
 energy generated =  14465101454699264.                                                                                                                                                                      
```

The error code `istate = 2` indicates that VODE thinks it was successful. In commit 9d3ddb13b8abf8953756ef6dc036ed569d27664b, I added some checks for the integration to fail in a case like this where X < 0 for one of the species, despite VODE's success code.

I think that perhaps this is too restrictive. In a case like this, the species that became negative was already fairly unimportant to begin with; it started at X = 4e-3. In this case, I believe we could get away with a simple flooring and renormalization without distorting the simulation too much.

The question is, how do we algorithmically separate out this case from the case where, say, a C12 abundance starting at X = 0.47 becomes negative in a single step, in which case we definitely do want to throw a failure?

maxpkatz:
Closing, I don't think I want to do anything about this. I'm OK with just paying the retry cost.

PR #350, opened 2020-07-20
VODE steps fail if the state is not valid

maxpkatz:
Commit 9d3ddb13b8abf8953756ef6dc036ed569d27664b added intentional failure modes to the VODE integrator to prevent VODE from giving us a state where the integrator believed it was successful but the state was not physically valid. Specifically, we checked on T < 0 and X < -0.01 or X > 1.01. Although this gave us the opportunity to then retry the burn, retries are expensive and we'd like to avoid them if possible (#106). One way we can significantly mitigate this is by preventing these failure modes from occurring at the end of each step rather than only at the end of the burn. We can then rely on the integrator's built-in step retry mechanism (which decreases dt or changes the integration order).

I chose to use the same failure modes as the one at the end of the burn. Additionally, I added a failure mode if any X changes by more than a factor of 2. The abundance never changes that much in a single step in a typical burn, but I have witnessed some extreme cases where the abundance floors in a single integration step and VODE doesn't reject the update. (I am not sure why.)

### DIFF
--- a/integration/VODE/vode_type.H
+++ b/integration/VODE/vode_type.H
@@ -22,6 +22,18 @@ const amrex::Real CCMXJ = 0.2e0_rt;
 const amrex::Real HMIN = 0.0_rt;
 const amrex::Real HMXI = 0.0_rt;
 
+// We will use this parameter to determine if a given species abundance
+// is unreasonably small or large (each X must satisfy
+// -failure_tolerance <= X <= 1.0 + failure_tolerance).
+const Real vode_failure_tolerance = 1.e-2_rt;
+
+// For each species whose abundance is above a certain threshold, we do
+// not allow its mass fraction to change by more than a certain amount
+// in any integration step.
+const Real vode_increase_change_factor = 2.0_rt;
+const Real vode_decrease_change_factor = 0.5_rt;
+const Real vode_change_factor_X_threshold = 1.0e-6_rt;
+
 // For the backward differentiation formula (BDF) integration
 // the maximum order should be no greater than 5.
 const int VODE_MAXORD = 5;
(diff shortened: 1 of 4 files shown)


integration/VODE/vode_type.H
```diff
@@ -22,6 +22,18 @@ const amrex::Real CCMXJ = 0.2e0_rt;
...
+// is unreasonably small or large (each X must satisfy
+// -failure_tolerance <= X <= 1.0 + failure_tolerance).
+const Real vode_failure_tolerance = 1.e-2_rt;
+
+// For each species whose abundance is above a certain threshold, we do
+// not allow its mass fraction to change by more than a certain amount
+// in any integration step.
+const Real vode_increase_change_factor = 2.0_rt;
+const Real vode_decrease_change_factor = 0.5_rt;
+const Real vode_change_factor_X_threshold = 1.0e-6_rt;
```
zingale, line 35: I think vode_change_factor_X_threshold should be max(1.e-6, atol_spec)

maxpkatz: Maybe I should make it atol_spec?

zingale: that's fine too

maxpkatz: Done

zingale:
can you update the CHANGES.md?

zingale:
all tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-07-20-001/index.html
so it seems we don't have any coverage for this change in the suite?

maxpkatz:
> all tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-07-20-001/index.html
> so it seems we don't have any coverage for this change in the suite?

As far as I can tell, this doesn't affect normal burns, it just affects burns that would have failed anyway with (say) X > 1. By construction, we didn't have this in the suite, because all our burns in the suite pass :-)

zingale:
okay, I'm happy with this one the CHANGES.md is updated

maxpkatz:
> can you update the CHANGES.md?

Done

PR #354, opened 2020-07-22
switch to using bion instead of ebin

zingale:
this is consistent with the aprox nets and eliminates a multiply on the sum over dY/dt

### DIFF
--- a/networks/triple_alpha_plus_cago/actual_network.F90
+++ b/networks/triple_alpha_plus_cago/actual_network.F90
@@ -2,6 +2,7 @@ module actual_network
 
   use network_properties
   use amrex_fort_module, only : rt => amrex_real
+  use fundamental_constants_module
 
   implicit none
 
@@ -28,6 +29,8 @@ module actual_network
 
   character (len=10), save :: reac_names(nrates)
 
+  real, parameter :: conv_factor = MeV2eV * ev2erg * n_A
+
 contains
 
   subroutine actual_network_init()
@@ -40,7 +43,7 @@ contains
     ! The binding energy per unit mass (erg / g) is obtained by converting
     ! the energies in MeV to erg then multiplying by (N_A / aion) where
     ! N_A = 6.0221415e23 is Avogadro's number
-    bion(ihe4)  = 28.39603_rt ! MeV / nucleus
+    bion(ihe4)  = 28.29603_rt ! MeV / nucleus
     bion(ic12)  = 92.16294_rt ! MeV / nucleus
     bion(io16)  = 127.62093_rt ! MeV / nucleus
     bion(ife56) = 492.25389_rt ! MeV / nucleus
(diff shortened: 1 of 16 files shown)


zingale:
it looks like if I convert the ebin value in the old form to MeV / nucleus I get 127.62077 instead of 127.62093, but I am not sure where that first value comes from.  The latter value matches what is in aprox21.

networks/triple_alpha_plus_cago/actual_network.F90
```diff
@@ -28,28 +29,30 @@ module actual_network
...
 contains
 
   subroutine actual_network_init()
 
     call network_properties_init()
-    allocate(ebin(nspec))
+    allocate(bion(nspec))
 
 
     ! our convention is that binding energy is negative.  The following are
```
maxpkatz, line 42: This comment needs to be updated

networks/triple_alpha_plus_cago/actual_network_data.cpp
```diff
@@ -20,10 +20,10 @@ void actual_network_init ()
     // the binding energies per unit mass (erg / g) obtained by converting
```
maxpkatz, line 20: This comment needs to be updated

networks/triple_alpha_plus_cago/actual_network.F90
zingale: comment updated

networks/triple_alpha_plus_cago/actual_network_data.cpp
zingale: I've updated this

zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-09-18-001/index.html

networks/triple_alpha_plus_cago/actual_network.F90
```diff
@@ -28,28 +29,30 @@ module actual_network
...
     ! the energies in MeV to erg then multiplying by (N_A / aion) where
     ! N_A = 6.0221415e23 is Avogadro's number
-    ebin(ihe4)  = -6.8253797e18_rt    !  28.39603 MeV / nucleon
-    ebin(ic12)  = -7.4103097e18_rt    !  92.16294 MeV / nucleon
-    ebin(io16)  = -7.6959581e18_rt    ! 127.62093 MeV / nucleon
-    ebin(ife56) = -8.4813001e18_rt    ! 492.25389 MeV / nucleon
+    bion(ihe4)  = 28.29603_rt ! MeV / nucleus
+    bion(ic12)  = 92.16294_rt ! MeV / nucleus
+    bion(io16)  = 127.62093_rt ! MeV / nucleus
+    bion(ife56) = 492.25389_rt ! MeV / nucleus
```
harpolea, line 49: aprox21 is using `bion(ife56) = 492.2450e0_rt`. Which value do we believe is correct?

maxpkatz: Let's sort that out after this PR. Eventually we should unify all of the physical constants across all the networks, but we don't have to tackle that as part of this code change.

maxpkatz (approved):

PR #358, opened 2020-07-24
C++ doesn't like 0-length arrays, so don't include aux in the eos_t if naux = 0

zingale:


zingale:
note: Castro and MAESTROeX will need to ifdef if out similarly

PR #363, opened 2020-08-01
split the integrators off into their own docs

zingale:


sphinx_docs/source/networks.rst
```diff
@@ -38,11 +38,13 @@ At a minimum, a network needs to provide:
...
 Most of these quantities are Fortran parameters.
 
-**A convention adopted in Microphysics is that each network
-is responsible for determining the energy release from a change
-in composition**. Most networks will provide an array of the species
-binding energies and a routine to compute the energy yield from
-the reaction rates.
+.. note::
+
+   A convention adopted in Microphysics is that each network is
```
harpolea, line 43: This appears to have lost the bold `**` from the previous version at the start of the sentence, but not at the end

zingale: fixed

PR #367, opened 2020-08-03
Github action for building docs

harpolea:
This removes the travis workflow for building the docs, and replaces it with a Github action which builds docs for the development and main branches (as done in Castro and MAESTROeX).


harpolea:
I just noticed that this also has the `extern_parameters.H` include statement in `burn_type.H` that I needed for the fully C++ version of MAESTROeX to build

PR #368, opened 2020-08-03
get some of the scattered unit tests building again

zingale:
A lot of networks have a test-cpp/ subdirectory that just exercises the RHS for that network.  This gets some of these building again, and removes all the dependencies on Castro.


dwillcox:
These tests all compile and run.

PR #369, opened 2020-08-04
NSE test problem

zingale:
add a test problem that reaches into NSE
    this is, in part, achieved by lowering the threshold composition mass fractions


zingale:
blocked by #332

zingale:
going to add this directly to one of the other PRs for testing

PR #370, opened 2020-08-05
move the naux info into a separate include file

zingale:
this will allow us to do make logic, like checking on whether USE_NSE is set.


zingale:
all tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-08-05-001/index.html

zingale:
all tests still pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-08-07-001/index.html

PR #371, opened 2020-08-06
Automatically link pynucastro electroncapture and betadecay files

ChrisDeGrendele:
When pynucastro generates rate files the user must copy them into the directory by themselves. This will do that automatically

This removes the URCA table which was a special case of this. 

I'm not sure if there are others besides `*capture.dat` and `*decay.dat` that can be generated?

I wanted to avoid blindly linking any `*.dat` because obviously there can be other things a user might put in there?

One question I had was in line 35/28 it says: "do not check to see if the file already exists, just force-link it in", which makes sense because something might have changed. But doesn't this still check to see if the file still exists?

Should all of them be replaced with something like? : 
```
echo Linking pynucastro rate file;  ln -sf $(PYNUCASTRO_FILE)
```
where the extra -f is for force.


zingale:
we should probably add a rule to delete all of these links in `clean` as well.

ChrisDeGrendele:
This works for me now. It automatically links and cleans `*decay.dat`, `*capture.dat`, `reaclib_rate_metadata.dat`, `helm_table.dat`, and `partition_function_table.dat` . Tested using MAESTROeX and Castro. 

(This PR also includes the partition function table which will be generated via https://github.com/pynucastro/pynucastro/pull/106 so maybe it'd be good to wait until that is merged? It runs as usual without the file though, so up to you.)

zingale:
this looks good, but I am a bit confused.  When I build `test_react` with the URCA-simple network and then do `make clean`, I get:
```
Cleaning...
rm -f tmp_build_dir/f/3d.gnu.EXE/*.f tmp_build_dir/f/3d.gnu.EXE/*.f90 tmp_build_dir/f/3d.gnu.EXE/*.F90
rm -f tmp_build_dir/s/3d.gnu.EXE/* tmp_build_dir/o/3d.gnu.EXE/*.o tmp_build_dir/o/3d.gnu.EXE/*.mod
rm -f tmp_build_dir/t/3d.gnu.EXE/*
rm -f *.optrpt tmp_build_dir/o/3d.gnu.EXE/*.optrpt
rm -f main3d.gnu.ex
rm -f -r main3d.gnu.ex.dSYM
rm -f extern.F90
rm -f network_properties.F90 network_properties.H
rm -f tmp_build_dir/microphysics_sources/3d.gnu.EXE/*.H tmp_build_dir/microphysics_sources/3d.gnu.EXE/*.[fF]90
rm -f extern_parameters.cpp extern_parameters_F.H extern_parameters.H
rm -f network.f90
rm -f extern.f90
rm -f *.dat
```

I am not sure where the `rm -f *.dat` is coming from -- that seems to not be your new rule

zingale:
oh... it is in `test_react/GNUmakefile`!

zingale:
so can you remove the blanket `rm *.dat` from `unit_test/test_react/GNUmakefile` and `unit_test/test_rhs/GNUmakefile`?

ChrisDeGrendele:
Good catch, I also noticed `helm_table.dat` is in the `clean` of MAESTROeX and Castro. Since both of their make files include Microphysics_extern, these will now be duplicates. Since the helm_table is generated here, I think its best to have the rm here too.

zingale:
this works, but 2 things:
1. because the clean rules start with `@`, we don't see the explicit `rm` output to the screen.  I kinda like to see what it is removing
2. if we build as, `make NETWORK_DIR=ignition_reaclib/URCA-simple` and then do `make clean`, we don't remove all of the electron capture files.  This is because the `GNUmakefile` defaults to a different network, and the logic doesn't pick them up.  We instead need to do `make NETWORK_DIR=ignition_reaclib/URCA-simple clean`.  Not sure if I like that, and maybe the blanket `rm *.dat` was better?

ChrisDeGrendele:
Oh, that is a good point. I was just wary of having a blanket `rm -f *.dat` show up in all of MAESTROeX and Castro because it could lead to accidental deletion of files. What if instead, we changed everything to have something like `*.network.dat`? So `helm_table.dat` for example would be `helm_table.network.dat`?

zingale:
I like that idea.  But I think we can leave helm_table.dat unchanged (since some of our workflow expects a file with that exact name) and just focus on naming the network tables something like *.network.dat.  Then we have an explicit rule to clean helm_table.dat and a separate rule to clean *.network.dat

Make.Microphysics_extern
```diff
@@ -18,24 +18,20 @@ ifndef GENERAL_NET_INPUTS
...
+PARTITION_FILE := "$(NETWORK_PATH)/partition_function_table.dat"
 all: pynucastrorates
 
 
 # do not check to see if the file already exists, just force-link it in
 # because different Reaclib networks will have different versions of the file
 pynucastrorates:
 	@if [ -f $(PYNUCASTRO_FILE) ] && [ ! -f ./reaclib_rate_metadata.dat ]; then echo Linking pynucastro rate file; ln -s $(PYNUCASTRO_FILE) .; fi
+	@if [ -f $(PARTITION_FILE) ] && [ ! -f ./partition_function_table.dat ]; then echo Linking pynucastro Partition Function Table; ln -s $(PARTITION_FILE) .; fi
+	$(@foreach file,$(TAB_FILES), if [ -f "$(file)" ] && [ ! -f ./$(notdir $(file)) ]; then ln -s $(file) . ; fi ;)
```
zingale, line 34: the `@` is in the wrong place here, so this doesn't work.  It should be
```
@$(foreach file,...)
```
and not
```
$(@foreach file,...)
```

zingale:
this seems to work now, but we need to modify the networks to look for the new names, e.g., `23Na-23Ne_electroncapture.network.dat` instead of `23Na-23Ne_electroncapture.dat`

Right now, things crash because it can't find those file names

Also, because we did this rename, the cleaning can be network invariant as 
```
$(RM) *network.dat
```
instead of
```
$(foreach file,$(TAB_FILES), if [ -L $(notdir $(file)) ]; then rm -f $(notdir $(file)); fi;)
```

zingale:
can you merge `development` into the PR so I can run the test suite on it?

zingale:
this is done now

PR #373, opened 2020-08-24
Warning fixes

zingale:
The static variables in `network_properties.H` are warned as unused with GCC everywhere we include the header and don't use them all.  This uses the `[[gnu::unused]]` attribute to silence these warnings.

### DIFF
--- a/interfaces/network.H
+++ b/interfaces/network.H
@@ -6,7 +6,9 @@
 
 void network_init();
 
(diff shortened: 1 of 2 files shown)


zingale:
GPU tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gpu/2020-08-24-001/index.html

PR #379, opened 2020-08-28
Updates the python wrappers

dwillcox:
This PR updates the python wrappers for Microphysics since some of the API has changed.

Main changes:

- The right hand side array `ydot` and jacobian `jac` are now stored as plain numpy arrays in the python `burn_state` object.

- A function is provided to transform `ydot` and `jac` from units of molar fractions into mass fractions.


harpolea (approved):

zingale:
closing this -- we no longer support the python wrappers, but most of what we used them for can be done with pynucastro now.

PR #380, opened 2020-08-30
update the extern_probin.template

zingale:
this adds the ability to output the runtime parameters to a job_info
and deletes an unused template


interfaces/extern_probin.template
```diff
@@ -9,10 +9,10 @@ module extern_probin_module
 
   private
 
-  @@declarationsA@@
+  @@declarations@@
```
ChrisDeGrendele, line 12: I'm trying to understand why this was like this in the first place? It looks like the python script that searches for this `@@` keywords looks for keywords, one of them being `declarations` in which case this was a mistake to have the A there? 

If that's indeed the case should `integration/VODE/cuVODE/test/extern_probin.template` Line 12 and 15 be changed to this as well? It also has `@@declarationsA@@` and `@@cudaattributesA@@`

zingale: it used to be that we had multiple different declaration regions, so we had a declarationsA and declarationsB.  I just never updated this.

PR #381, opened 2020-08-30
remove the burn_to_vode in jac() -- it appears unnecessary

### LINKED ISSUE #364: is burn_to_vode needed in vode_rhs.H for the Jacobian
Evaluating the Jacobian does not update the integrated state, so I don't think that we need to call ``burn_to_vode`` at the end of ``jac()`` in ``vode_rhs.H``

zingale:
closes #364


zingale:
while all tests pass, it seems that none of the C++ tests exercise the analytic Jacobian.  Need to add some tests that do this first, and then retest this PR.

PR #385, opened 2020-09-01
Update extern_probin.template

ChrisDeGrendele:
These were never updated either. See PR #380

### DIFF
--- a/integration/VODE/cuVODE/test/extern_probin.template
+++ b/integration/VODE/cuVODE/test/extern_probin.template
@@ -9,10 +9,10 @@ module extern_probin_module
 
   private
 
-  @@declarationsA@@
+  @@declarations@@
 
 #ifdef AMREX_USE_CUDA
-  @@cudaattributesA@@
+  @@cudaattributes@@
 #endif
 
 end module extern_probin_module


maxpkatz:
Does this test currently work? We don't often exercise it (though I did look at it a bit when I did the C++ port).

zingale:
oh... now I understand things.  We are still using the version of the script in `AMREX_HOME/Tools/F_scripts/`:
```
$(MICROPHYSICS_AUTO_SOURCE_DIR)/extern.F90: $(EXTERN_PARAMETERS) $(EXTERN_TEMPLATE)
	@if [ ! -d $(MICROPHYSICS_AUTO_SOURCE_DIR) ]; then mkdir -p $(MICROPHYSICS_AUTO_SOURCE_DIR); fi
	$(AMREX_HOME)/Tools/F_scripts/write_probin.py \
           -t $(EXTERN_TEMPLATE) -o $(MICROPHYSICS_AUTO_SOURCE_DIR)/extern.F90 -n extern \
           --cxx_prefix $(MICROPHYSICS_AUTO_SOURCE_DIR)/extern \
           --pa "$(EXTERN_PARAMETERS)" $(MANAGED_PROBIN_FLAG)
```
but we should switch over to the version in `Microphysics/util/build_scripts/`

In fact, Castro and MAESTROeX have already switched over to that.

zingale:
I'm not sure what "test" you mean, but the extern parameter stuff does work for unit tests.  I'm realizing that there is some redundancy in our build system stuff that I will try to clean up shortly.

ChrisDeGrendele:
Ah, that makes sense now. @zingale

maxpkatz:
> I'm not sure what "test" you mean

The unit test that is being edited here (the standalone cuVODE driver).

zingale:
oh, indeed, I have no idea about that test.

zingale:
I don't think this is relevant anymore

PR #386, opened 2020-09-01
switch to local copies of the build scripts

zingale:


zingale:
note: managed is the default now

PR #390, opened 2020-09-02
microphysics_math / esum is only needed for reactions

zingale:


zingale:
all tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-09-02-001/index.html

PR #391, opened 2020-09-05
add versions of the init / finalize callable from C++

zingale:


interfaces/microphysics.F90
```diff
@@ -18,10 +18,19 @@ module microphysics_module
 
 contains
 
+  subroutine microphysics_init_C(small_temp, small_dens) bind(C, name="microphysics_init_C")
```
maxpkatz, line 21: This interface is confusing, I think anything with a `_C` suffix should be initializing the C++ version of the code, not be a Fortran interface.

zingale: renamed to `microphysics_initialize()`

maxpkatz (approved):

PR #393, opened 2020-09-11
remove simplified-SDC evolution from the BS integrator

zingale:
this is not tested, not maintained, and not used
it also doesn't have a GPU path


zingale:
all tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-09-11-002/index.html

maxpkatz (approved):

PR #394, opened 2020-09-13
create a new YdotNetArray1D for interfacing with the actual_rhs interface

zingale:
For both Strang and SDC, this is what we will come in with.  For Strang, this
array is the same size as the system we are integrating, but that is not the
case for SDC (neqs != VODE_NEQS), so a clear data type here helps keep things straight.


maxpkatz:
I prefer that we template on the Array1D size instead of using this typedef.

zingale:
that's fine, but I'll still need this type for the internal bits in SDC

zingale:
thinking about this more, the advantage of using this typedef is that we _always_ need to call the network's RHS with this size array -- there is no flexibility built into the network rhs routine.  So this just makes a convenient name to use to declare something of this size.  For SDC this is important, because the ydot that VODE carries is different than the ydot that the network cares about, so we need to create a temporary array, the YdotNetArray1D, to call the network stuff.

PR #362, opened 2020-07-31
Create a C++ version of aprox19 + NSE

zingale:
This mirrors the Fortran aprox19 + NSE network

PR #398, opened 2020-09-19
switch to use the same C++ variables class as other tests

zingale:


zingale:
this extracts these changes from #362, since they are independent of that change.

zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-09-19-005/index.html

maxpkatz (approved):

PR #404, opened 2020-10-16
make the Strang NSE update more thermodynamically consistent

zingale:
This uses a method slightly different than the Ma et al. 2013 update for the NSE.  Instead of just taking Ye from the first call to the NSE table, we instead predict what the new Ye will be and re-call the table to get consistent composition and energy release from the table.


zingale:
This now updates both the Fortran and C++ code paths.  This change has been shown in the SDC paper to give better convergence with Strang + NSE.

maxpkatz (approved):

PR #413, opened 2020-10-31
Implement iso7 RHS terms in a standardized form using data from a constexpr function

maxpkatz:
This PR is the start of a series of refactors so that iso7's RHS is re-implemented by moving all of the RHS data to the header in a function called `rhs_data` that returns an `rhs_t` struct. Every term in an RHS is a multiplicative combination of up to five terms: a prefactor, `Y(i)`, `Y(j)`, `Y(k)`, and `rate(l)`. Most reactions have two `Y` factors, like C12 + C12 -> Mg24; some have only one `Y` factor, like C12 + gamma -> 3He4; and the three-body reactions (triple alpha) can have three. If the species index is -1, we know to ignore it. If the compiler can optimize this suitably well by evaluating these terms at compile time, this should add no additional runtime cost. This initial change will do the evaluation at run time and incur a performance cost, but that will be resolved in a future PR.

The motivation for this change is that we want the flexibility to experiment with integrators that can directly utilize information about the form of the RHS. Right now the RHS is a black box to the integrator, but if we implement every RHS in a standard form like this, then we can write general integrators that can perform more sophisticated integration algorithms (but that can still construct the standard form of the RHS if needed, since they have all the terms ready to do so).


maxpkatz:
Feedback on the structure is welcome -- at any rate, it's plausible it could change in the future as I use this to test out other integration schemes.

networks/iso7/actual_rhs.H
```diff
@@ -671,7 +671,54 @@ void ener_gener_rate(T const& dydt, Real& enuc)
 
 }
 
+template<int spec, int j>
+AMREX_GPU_HOST_DEVICE AMREX_INLINE
+Real rhs_term(Array1D<Real, 1, NumSpec> const& y, rate_t const& rr)
```
maxpkatz, line 676: Later I will hide these functions in one of the headers in interfaces/ because it's supposed to be general and that way we won't need to have the implementation bits in the networks. But for now this is the only network implementing this so I'm leaving it here.

zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-11-01-002/test_react_C-iso7-VODE.html

maxpkatz:
@dwillcox do you have any feedback on this idea?

dwillcox:
I haven't yet done an actual review and checked all the rates, but I do like the approach.

It seems really fancy, and I like that it abstracts away the code pattern for building the RHS terms.

The nested case statements are somewhat cumbersome to read, and I might suggest using enums to label the rates in each case statement to improve readability.

maxpkatz:
A way that we could clean this up quite a bit is, rather than using the contrived RHS "terms," iterate over all rates and add their contribution to the RHS. Rates that don't contribute to a given RHS would just be zeroed out. Then `rhs_data` would be a simple function of just the species and the rate, and as a result every species could be treated uniformly and we could apply some tricks to avoid the nested switch statements (Don pointed me to https://www.fluentcpp.com/2017/06/27/how-to-collapse-nested-switch-statements/). But what I need to solve there is that we still want to construct the array of non-zero terms so that we can apply `esum` to it, which means I need to figure out at compile time how to construct that minimally sized array rather than an array sized to the number of rates which is mostly zero for any given term. (It occurs to me also that we could apply esum inline as we accrue rates one by one, and if we calculated the terms using constexpr functions maybe we could compile out the zero terms, but that's an optimization for later.)

maxpkatz:
I have updated the scheme a bit based partially on Don's feedback, which was aligned with some other changes I wanted to make. From the network's perspective, I've implemented the simplification that we can have a single switch statement instead of nested switch statements (using a simple approach of automatically assigning a unique numerical identifier to each species x rate combination). From the implementation perspective, I'm constructing the RHS by looping over rates, identifying the rates that contribute to the RHS for each species, and then filling the array which we then pass to esum.

With this infrastructure in place, and because we now fully construct the RHS behind the scenes, we should be able to do any number of things to experiment with integration schemes. For example, we could look into the idea I mentioned of constructing the RHS as a running esum rather than explicitly constructing the array and only then applying esum to it. We could also experiment with constructing the rates one by one and filling the RHS as we go, rather than explicitly constructing the full rate array. More generally I want to experiment with other integration strategies which require access to direct information about the structure of the RHS terms.

Ready for review.

maxpkatz:
I've greatly simplified this PR to add review. The basic idea is still there, but I've only applied it to one term in iso7 -- the Ni56 RHS -- and for now we're doing the evaluation at run time, not compile time. This will temporarily slow down iso7 but I have been able to demonstrate that when the changes are done, the performance will be recovered.

zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-11-19-002/index.html

zingale (approved):

PR #416, opened 2020-10-31
Remove array over NumGroups from rate_t

maxpkatz:
In every place in the networks where we are using rate_t, it works just as well to have an array of rate_t, where the size of the array is the number of rate groups, as it does to have the rate_t be a 2D array. Making this change has two benefits: we can clean up some of the RHS code later, because the RHS (unlike the Jacobian) typically does not actually need to fill all the rate groups (so this will save work and memory), and also this will support the change in #413.

### DIFF
--- a/networks/aprox13/actual_rhs.H
+++ b/networks/aprox13/actual_rhs.H
@@ -39,7 +39,7 @@ namespace RateTable
 
 
 AMREX_GPU_HOST_DEVICE AMREX_INLINE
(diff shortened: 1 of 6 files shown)


maxpkatz:
I've only done this in C++, doesn't seem worth the effort to do the same to Fortran.

zingale (approved):

zingale:
all tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-10-31-001/index.html

PR #417, opened 2020-11-01
Guard against pulling in actual_network

maxpkatz:
We pull in this file if USE_CXX_REACTIONS = TRUE, but if that is set without USE_REACT = TRUE, we will attempt to pull in actual_network.H, which only exists for some networks.


zingale (approved):

zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-11-01/index.html

PR #418, opened 2020-11-01
Continue iso7 cleanup

maxpkatz:
Continuing the idea from #413, we refactor the iso7 RHS to look like something that could be standardized across networks. actual_rhs now looks like:

```
void actual_rhs(burn_t& state, Array1D<Real, 1, neqs>& ydot)
{
    // Evaluate the rates
    
    rate_t rr;
    evaluate_rates(state, rr);

    // Call the RHS to get the species terms

    spec_rhs(state, rr, ydot);

    // Instantaneous energy generation rate

    ener_rhs(state, ydot);

    // Append the temperature equation

    temperature_rhs(state, ydot);
}
```

The idea here is that in this form, the individual network bits can be called from outside the network itself, if the integrator chooses to do so. Eventually we may not even need to explicitly construct the full RHS in the network itself, if the network provides all of these individual components.

Since the functions are designed to be called from outside the network, we standardize on always passing around the burn_t itself rather than pieces of it (which is likely to be non-uniform across networks). This necessitates two other changes. First, we write a new version of the rates routine which will only work from the RHS right now (because it accepts a scalar rate_t rather than an array). Second, the energy generation rate routine (now called `ener_rhs`) also includes the neutrino cooling directly (and we leave the Jacobian out of that change for now, it continues to use the old interface).


maxpkatz:
Closing -- this will be done as part of the series of smaller refactors (#413, #433, etc.).

PR #419, opened 2020-11-01
Add interface for C++ integrator to use Fortran RHS

maxpkatz:
If a network has a C++ RHS implementation, it should set `DEFINES += -DNETWORK_HAS_CXX_IMPLEMENTATION` in its Make.package, and the C++ integrator will continue to use it. Otherwise, we default to the Fortran RHS, and we add a wrapper script that translates between the C++ RHS/Jac and the Fortran RHS/Jac.

With this change, any downstream application can use `USE_CXX_REACTIONS = TRUE` everywhere and still use any network. The only catch is that there is no CUDA implementation right now in this PR, so the legacy Fortran networks won't run on GPUs. If we need to fix that later, it should be possible.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,8 +2,8 @@
 
    * An interface has been added for C++ integrators to call the RHS
      from a network that only has a Fortran implementation. This allows
-     the use USE_CXX_REACTIONS = TRUE for any network (however, CUDA is
-     not currently supported for this case). (#419)
+     the use of USE_CXX_REACTIONS = TRUE for any network (however, CUDA
+     is not currently supported for this case). (#419)
 
 # 20.11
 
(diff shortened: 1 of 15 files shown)


maxpkatz:
I have tested this with test_rhs for aprox21 (its C++ implementation has not been merged yet) and USE_CXX_REACTIONS = TRUE and I verify a diff of exactly zero with do_cxx = 1 and do_cxx = 0, as expected. I've also tested Castro rotating_star using aprox21 with and without USE_CXX_REACTIONS = TRUE and see the expected diff of O(~1e-5) compared to the Fortran integrator.

zingale (approved):

zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-11-01-007/index.html

zingale:
can you add this to the CHANGES?

maxpkatz:
> 
> 
> can you add this to the CHANGES?

Done.

PR #423, opened 2020-11-10
start porting NSE to SDC

zingale:
This puts the auxiliary data into the SDC code path and syncs it up with the mass fractions as needed.  This will allow us to get the thermodynamic information from aux instead of X.


zingale:
note: review with whitespace disabled.

zingale:
Here's the current status of this PR:

We want to be able to run simplified SDC with NSE.  To do this, we need to do 2 things:
1. modify the C++ simplified-SDC code paths to work with the auxiliary data for thermodynamics
2. write the SDC update for when we are using the NSE table.

This PR addresses the first point.

When we compile the code with `USE_NSE=TRUE`, this switches the equation of state to always take the composition from the auxiliary data instead of the mass fractions.  This switch is done in `eos_composition.H`.  When we are integrating the reaction network directly instead of using the NSE table, then we actually care about the mass fractions, X.  So anywhere in the simplfiied SDC VODE code path where we call the EOS, we need to make sure that the auxiliary data is in sync with the mass fractions.  That's what we try to address here.

If we have everything properly synced, then we can run the Castro ``reacting_convergence`` test problem with and without `USE_NSE` to compare the case where the EOS uses the mass fractions vs. the case where the EOS uses the auxiliary data, and they should agree (and both converge 2nd order).

At the moment, they do not agree -- the `USE_NSE=TRUE` SDC run does not get 2nd order convergence in T and Abar.  Here's how to see this.  

In `reacting_convergence/`, edit the script `convergence_simplified_sdc.sh` and change `DIM=2` to `DIM=1`.  You can also cut out the second block on tests which test 3 SDC iterations.

1. mass fraction case:

compile as:
```
make NETWORK_DIR=aprox19 USE_SIMPLIFIED_SDC=TRUE DIM=1 -j 20
```
then run the script:
```
./convergence_simplified_sdc.sh
```
There are 2 output files showing the convergence, `convergence.1d.lo.simplified_sdc_iter2.out` and `convergence.1d.hi.simplified_sdc_iter2.out`.  Look at the L1 errors in either file (L0 come first), you should see 2nd order convergence in the results. 

2. auxiliary data case:

compile as:
```
make NETWORK_DIR=aprox19 USE_SIMPLIFIED_SDC=TRUE DIM=1 -j 20 USE_NSE=TRUE
```
and rerun the script:
```
./convergence_simplified_sdc.sh
```
Now in the convergence output files, you'll see that T and Abar are first order, something like:
```
Level  L1 norm of Error in Each Component
-----------------------------------------------
level: 0
                 density        7.735466e+10       2.532693   1.336811e+10
                    xmom        3.846154e+19       2.624164   6.238419e+18
                    ymom        0.000000e+00   ------------   0.000000e+00
                    zmom        0.000000e+00   ------------   0.000000e+00
                   rho_E        2.285265e+28       2.562136   3.869515e+27
                   rho_e        1.994370e+28       2.567164   3.365212e+27
                    Temp        2.970279e+13       1.207740   1.285971e+13
```

zingale:
update: with PR #491, I get perfect agreement in the `test_sdc` unit test comparing `USE_NSE=TRUE` and `USE_NSE=FALSE`

zingale:
I think that this is ready to review.  Based on the tests done with `test_sdc` and the new unit test embedded in here, I believe that this works as expected.  There is still the bit needed about actually doing the SDC update if we do the NSE table (that'll be in another PR) and there looks like there is an issue in Castro to be sorted out.

harpolea (approved):

integration/VODE/vode_type_simplified_sdc.H
```diff
@@ -73,9 +77,10 @@ void sdc_burn_to_eos (const burn_t& state, T& eos_state)
 
 #if NAUX_NET > 0
     // aux
-    for (int n = 0l n < NumAux; n++) {
+    for (int n = 0; n < NumAux; n++) {
         eos_state.aux[n] = state.y[SFX+n] / eos_state.rho;
     }
+    //set_nse_aux_from_X(eos_state);
```
maxpkatz, line 83: Was the intention for this to remain commented?

zingale: yes, I should remove that

PR #424, opened 2020-11-14
Allow other C++ integrators

maxpkatz:
This syncs up with the Fortran scheme where (at least for now) we'll have exactly one actual_integrator.H per build, which allows us to easily swap them out when we add others.


zingale:
should the SDC C++ VODE integrator also be renamed?

maxpkatz:
Yes, good point. Updated.

zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-11-14-002/index.html

zingale (approved):

PR #433, opened 2020-11-19
iso7 part 2

maxpkatz:
Continues #413 by extending the approach to all species

PR #434, opened 2020-11-19
iso7 part 3

maxpkatz:
Continues #433 by looping over species in addition to looping over rates. In order to make the loop work we resize the esum array to the number of total rates (since Array1D's size needs to be known at compile time). This means that esum will have some redundant work -- this de-optimization will eventually be recovered.


zingale (approved):

zingale:
test: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-12-04-004/test_react_C-iso7-VODE.html

PR #439, opened 2020-11-21
iso7 part 5

maxpkatz:
A version of the rate evaluation is added which works only on a single rate group (so only makes sense to call from actual_rhs).


zingale (approved):

zingale:
test: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-12-04-007/index.html

PR #441, opened 2020-11-22
iso7 part 6

maxpkatz:
A version of the energy equation is created just for the RHS (directly includes the neutrino term).


zingale:
tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-12-04-008/test_react_C-iso7-VODE.html

zingale (approved):

PR #442, opened 2020-11-22
iso7 part 7

maxpkatz:
Directly inline species_rhs into actual_rhs and actual_jac


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-12-04-009/test_react_C-iso7-VODE.html

zingale (approved):

PR #449, opened 2020-12-03
iso7 part 8

maxpkatz:
iso7 has a rate dependence on Y since the (7*He4 + Si28 -> Ni56) effective rate (both forward and reverse) depends on Y. The forward rate is given by k(rho, T) * Y(He)^3 *  Y(Si28), so several Jacobian elements need to have d(rate)/dY terms to be computed correctly. (A similar thing is true for the reverse rate.) The current code knows exactly which d(rate)/dY terms are computed and stores them in rate arrays dratedy1 and dratedy2. But if we want to generalize this we need to be agnostic to which specific rates and which specific species have the dependence. A solution is to allow the Jacobian to generally be a function of the derivatives of the rates with respect to Y in addition to the derivatives of Y itself in the RHS. To effect this generally we need to have a dratedY container which has NumSpec species and NumRates rates, and this change enables that. The vast majority of these entries will be empty, so this temporarily increases the storage requirements and adds some redundant calculations. But the change we'll be doing later is to detect the nonzero ones and only add them in at compile time.


zingale:
test: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-12-05-001/index.html

networks/iso7/actual_network.H
```diff
@@ -62,7 +62,7 @@ namespace Rates {
                        irni2si,
                        NumRates = irni2si};
 
-    const int NumGroups = 4;
+    const int NumGroups = 2 + NumSpec;
```
zingale, line 66: this doesn't seem to be used

networks/iso7/actual_rhs.H
```diff
@@ -377,8 +377,8 @@ void screen_iso7(const Real btemp, const Real bden,
        denom     = std::pow(bden * y(He4), 3.0e0_rt);
 
        rate(irsi2ni)     = yeff_ca40*denom*rate(ircaag)*y(Si28);
-       dratedy1(irsi2ni) = 3.0e0_rt * rate(irsi2ni)/y(He4);
-       dratedy2(irsi2ni) = yeff_ca40*denom*rate(ircaag);
+       dratedY(He4, irsi2ni) = 3.0e0_rt * rate(irsi2ni)/y(He4);
```
zingale, line 380: is this supposed to represent d(rsi2ni)/dHe4? the order of the arguments in the array make it seem like the opposite.

maxpkatz: Yes, that is what it represents.

zingale (approved):

PR #452, opened 2020-12-03
have test_sdc use C++ initialization

zingale:


zingale:
all test_sdc tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-12-03-006/index.html

maxpkatz (approved):

Issue #453, opened 2020-12-04
allow for Fortran building to be disabled


zingale:
We should start wrapping the EOS and Reaction stuff so that we can optionally skip compiling them.

We can do this either with separate `USE_F90_EOS`, `USE_F90_REACTIONS`, ... to enable/disable different parts or through a single `USE_F90_MICROPHYSICS`, to turn off all of F90 (eventually).

maxpkatz:
For reactions a consideration is that we have a mix of networks with C++ or Fortran support and so USE_F90_REACTIONS may have to be conditional on which network you're using (for now).

maxpkatz:
This now exists and is used in several tests. We cannot completely turn off Fortran yet because we still have some Fortran pynucastro networks, but after those are removed we can finish the Fortran elimination.

PR #454, opened 2020-12-04
Remove C++ interface to Fortran small_dens/small_temp

maxpkatz:
This interface is unnecessary because the C++ EOS initialization duplicates what the Fortran does regarding mintemp/mindens, and this is one of the blockers for #453.


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-12-05/index.html

zingale (approved):

PR #455, opened 2020-12-05
iso7 9/∞

maxpkatz:
Now we start applying the same approach to the Jacobian that we did for the RHS. We assert that the Jacobian terms have the same form as the RHS terms, K * rate * Y1 * Y2 * Y3 (although I think there cannot be any three-body terms). A complicating factor is that the rate derivative d(rate) / dY may be non-zero since many of the rates in the aprox networks have hidden contributions from equilibrium sequences that depend on Y. So in addition the Jacobian have a term from K * drate/dY * Y1 * Y2 * Y3.


zingale (approved):

zingale:
test: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-12-09-001/test_react_C-iso7-VODE.html

zingale:
do you expect changes in the answer?

maxpkatz:
> 
> 
> do you expect changes in the answer?

I expect roundoff changes in the RHS evaluation, yes. Now we're adding NumRates quantities with esum, where most entries are zero, whereas before we were only adding the non-zero terms. (We'll return to only adding the non-zero terms later using the C++14 stuff.)

PR #457, opened 2020-12-05
iso7 11/∞

maxpkatz:
Convert remaining Jacobian terms to new scheme


networks/iso7/actual_network.H
```diff
@@ -474,6 +474,228 @@ namespace RHS {
...
+            data.prefactor = -1.0_rt;
+
+            data.rate_prefactor = -1.0_rt;
+            data.rate_specindex1 = He4;
+            break;
+
+        case jac_rate<NumSpec, NumRates>(Si28, He4, irni2si):
+            data.rate_prefactor = 1.0_rt;
+            data.rate_specindex1 = Ni56;
+            break;
```
zingale, line 645: this is supposed to correspond to:
b(4) =  rr(2+He4).rates(irni2si) * y(Ni56);

is the indexing using 2+He4 automatic based on the species that appears in the denominator of the Jacobian element?

maxpkatz: `rr(2 + spec)` is d(rate)/d(spec), so the fact that this is d(Si28) / d(He4) implies that the relevant term is automatically d(rate) / d(He4).

zingale (approved):

zingale:
test suite fails to converge: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-12-12/test_react_C-iso7-VODE.html

maxpkatz:
> test suite fails to converge: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-12-12/test_react_C-iso7-VODE.html

This appears to be due to the fact that a couple of terms which were in single expressions were split up into multiple terms (resulting in different floating point order of operations).

PR #458, opened 2020-12-05
allow the EOS to be turned off by setting USE_FORT_EOS=FALSE

zingale:
By default, we set `USE_FORT_EOS ?= TRUE` in the `Make.Microphysics_extern`, so no codes will need to be changed.
This addresses part of #453


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-12-05-003/index.html

maxpkatz (approved):

PR #459, opened 2020-12-06
Pass small_temp and small_dens into eos_init

maxpkatz:
This is needed because the downstream codes do not make these variables available as globals without a namespace.


zingale:
all tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-12-06/index.html

zingale (approved):

PR #460, opened 2020-12-09
setup_composition does not need to be on the GPU

zingale:
this fixes GPU compilation


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gpu/2020-12-09-001/test_react-aprox13-VODE90.html
these errors are because of the conversion to C++ for the initialization

maxpkatz (approved):

PR #465, opened 2020-12-18
iso7 13/∞

maxpkatz:
species_rhs and ener_rhs return their values.


zingale:
all tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-12-18-002/index.html

PR #466, opened 2020-12-18
Add explicit integrator with asymptotic approximation

maxpkatz:
The explicit integration method of [Guidry et al. (2013)](https://ui.adsabs.harvard.edu/abs/2013CS%26D....6a5001G/abstract) (paper 1 of 3) is implemented. In this approach we detect which terms in the species RHS would equilibrate within the current integration timestep and then use the asymptotic approximation to update that term instead of normal explicit integration.

This seems to work OK for "easy" burns at low temperature -- the answer is fairly inaccurate for the default choice of parameters, presumably due to the first order method, but if we are very restrictive on the timestep then the answer converges to what VODE provides. Near NSE we just run into the same problems as the simple forward Euler integrator, which will require the "partial equilibrium" method that detects which specific rates are in approximate detailed balance and removes them from the integration.

To do:

- [ ] Use a higher-order integration method for both the explicit and asymptotic terms (the paper describes a predictor-corrector scheme)
- [ ] Add the partial equilibrium method from paper 3 of that series

### DIFF
--- a/integration/ExplicitAsymptotic/Make.package
+++ b/integration/ExplicitAsymptotic/Make.package
@@ -1,0 +1,1 @@
+CEXE_headers += actual_integrator.H
--- a/integration/ExplicitAsymptotic/README.md
+++ b/integration/ExplicitAsymptotic/README.md
@@ -1,0 +1,5 @@
+An explicit integrator using the asymptotic approximation proposed by
+M. Guidry et al. in Computational Science & Discovery 6 (2013) 015001,
+"Explicit integration of extremely stiff reaction networks: asymptotic methods."
+The timestepping is determined by not letting any integration
+quantity change by more than a given factor in a timestep.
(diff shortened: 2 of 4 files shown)


maxpkatz:
The predictor-corrector method for the asymptotic approximation is given in Equation 16 of the paper. It is not difficult to implement from a code perspective but I am unsure how to actually do it. The terms tau_p and F_p are based on the rates evaluated using the predictor. But both of these depend on the full Y vector to evaluate, so what Y is it? Is it the predictor Y with only the given term updated, which is simple to do but is an inconsistent state? Is it the Y with every value updated to the predictor, and if so what do we do about the terms that aren't being updated in the asymptotic approximation?

zingale (approved):

maxpkatz:
I think the next step here is to switch this over to the algorithm in the second paper that uses the quasi-steady-state (QSS) approximation. The terms are all evaluated using the same scheme, so that should resolve the trickiness associated with having one set being evolved using forward Euler and another set being evolved in the asymptotic approximation. The QSS method is a second-order predictor-corrector scheme so it should be more accurate.

PR #467, opened 2020-12-19
iso7 14/∞

maxpkatz:
Evaluate as much of the RHS as possible at compile time. Specifically we do not waste memory or flops on contributions to each RHS from rates that don't actually contribute.

### DIFF
--- a/interfaces/rhs_utilities.H
+++ b/interfaces/rhs_utilities.H
@@ -17,88 +17,6 @@ Array1D<Real, 1, NumSpec> species_rhs (const burn_t& state, rate_t const& rr);
 
 namespace rhs_impl {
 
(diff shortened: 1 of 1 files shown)


zingale:
there are some very large differences: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-12-19/test_react_C-iso7-VODE.html

maxpkatz:
Unfortunately there probably would be for the same reason we saw in #457.

maxpkatz:
To make this simpler to review I am retaining the old behavior that the esum is over NumRates terms even though most of them are zero. I will do the piece that skips the zero terms in a later PR.

zingale:
the latest tests look much better

zingale (approved):

PR #468, opened 2020-12-19
create a C++ numerical Jacobian

zingale:
This is not needed by VODE, since VODE internally does its own thing, but this could be useful for other integrators and also for the SDC implementations where we just do Newton iterations instead of an ODE solve.

This PR also introduces a unit test to test the C++ numerical Jacobian.


zingale:
This will help with https://github.com/AMReX-Astro/Castro/pull/1208

maxpkatz (approved):

PR #478, opened 2020-12-22
iso7 15/∞

maxpkatz:
Size the RHS array to be exactly the number of non-zero terms. This finally returns the actual work done to the original implementation.


zingale:
errors are reasonable: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-12-22-001/test_react_C-iso7-VODE.html

zingale (approved):

Issue #470, opened 2020-12-20
Template the EOS interface on eos_t variants and burn_t


maxpkatz:
We have had a couple issues in the past where we wanted to shrink the size of the eos_t for cases where some of the output data (like the derivatives) is unneeded (#202) and for cases where X doesn't need to explicitly be part of the eos_t, which is true for most equations of state (#206). Also, it would be nice to evaluate the EOS with a `burn_t`, rather than our hack of making a temporary `eos_t` and using the `burn_to_eos` and `eos_to_burn` interfaces. So ideally the EOS would be templated on all these types, and would work correctly regardless of outputs.

In C++17 we have `if constexpr`. What that allows us to do is write code like
```
template<class T>
void eos (int input, T& state)
{
    ...
    if constexpr (std::is_same<T, eos_t>::value) {
        state.dpdA = ...
    }
    ...
}
```
which will be conditionally compiled out if the input type is an object that is not an `eos_t` and thus doesn't have `dpdA`. If done correctly this eliminates the need for things like the `EXTRA_THERMO` ifdef. Possibly we will have something like an `eos_core_t` class that just contains a base set of variables, while the full `eos_t` has all of the variables including derivatives, and possibly is a derived class from `eos_core_t`. (Note that `std::is_same` is false when comparing a base class and a derived class.)

maxpkatz:
This will enable much more targeted EOS evaluations in general. For example, in the burner the only field we actually need right now is `c_V` or `c_P` (or just `T`, if we do #471). So based on the components of the input state, we could automatically detect which calculations in the EOS we need to do, and skip all the unnecessary ones.

maxpkatz:
This has already been done, closing.

PR #480, opened 2020-12-23
Template functions which accept an eos_t

maxpkatz:
Starts #470 by making all functions which previously accepted an eos_t be templated.


zingale (approved):

zingale:
tests are good: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-12-23-001/index.html

PR #481, opened 2020-12-23
Add types to determine if an EOS type has given members

maxpkatz:
To do #470 we will need to be able to determine at compile time if a given template type has certain member data. This adds a few structs for that purpose. The idea is that they can be used like

```
    if constexpr (has_xne_xnp<T>::value) {
        state.xne = xnefer;
        state.xnp = 0.0e0_rt;
    }
```

The intended usage above requires C++17, but this PR does not yet require that.

### DIFF
--- a/interfaces/eos_type.H
+++ b/interfaces/eos_type.H
@@ -53,6 +53,62 @@ struct eos_t {
 
 };
 
+template <typename T, typename Enable = void>
+struct has_energy
+    : std::false_type {};
+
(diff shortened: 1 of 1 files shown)


zingale:
tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-12-23-002/index.html

zingale (approved):

PR #485, opened 2020-12-24
Update to C++17 as minimum standard

maxpkatz:


zingale:
MAESTROeX and Castro now both require C++17

zingale (approved):

PR #492, opened 2020-12-28
remove support for F90 simplified SDC with VODE

zingale:
We are transitioning to C++ in the application codes that use SDC,
so the F90 simplified SDC interface is unused and will not get new
features.


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-12-28-001/index.html

I need to update test_sdc

zingale:
updated tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2020-12-29-003/index.html

maxpkatz (approved):

PR #495, opened 2020-12-29
Address various compiler warnings

maxpkatz:
- Use of single argument static_assert without a message is C++17-only
- One usage of static_assert in helmeos was prematurely implemented at compile time instead of runtime
- Shadowing of small_dens and small_temp in eos.H
- Unknown compiler pragmas (pragma gpu)
- Invalid use of compound literals in eos_type.H (C99 only)
- slen is unused in extern_parameters.cpp if there are no string parameters

### DIFF
--- a/EOS/breakout/actual_eos.H
+++ b/EOS/breakout/actual_eos.H
@@ -25,7 +25,7 @@ template <typename I>
 AMREX_GPU_HOST_DEVICE AMREX_INLINE
 bool is_input_valid (I input)
 {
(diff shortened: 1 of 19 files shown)


zingale (approved):

zingale:
all the C++ runs die with "amrex::Error::0::Unknown EOS input !!!"

maxpkatz:
> 
> 
> all the C++ runs die with "amrex::Error::0::Unknown EOS input !!!"

Should be fixed

PR #496, opened 2020-12-29
Do not integrate T; get it from e

### LINKED ISSUE #471: make call_eos_in_rhs permanently on and remove T integration
Right now, we solve: dX/dt, dT/dt, and de/dt.  The latter 2 are redundant.  We integrate e to accumulate the energy release over the burn (taking into account neutrino loses, etc).  And we integrate T to get an updated temperature to evaluate the rates during the subcycling of the integration.  Additionally, the reason for `call_eos_in_rhs` is to update the specific heats needed for the T integration.   But all that really does is do `eos(rho, T) -> c_v`.

A more consistent way to do things would be to just integrate dX/dt and de/dt and then always call the eos to get T from rho, e.  This will eliminate 1 equation from the ODE system and could make the overall integration easier.

Note:
* This only applies to Strang integration.  For the various types of SDC integration we are effectively already doing this.
* We would need to cast the analytic Jacobian in terms of e -- we can follow the model used for simplified SDC to translate it to the correct independent variables.  That works very well for simplified SDC.

maxpkatz:
Closes #471


zingale:
can you merge `development` into this?

zingale:
`test_react` with `aprox13` and `inputs_aprox13` fails with "too many steps" when run as-is.  But if you switch to the numerical Jacobian (via `integrator.jacobian=2`) then it completes.

maxpkatz:
#550 fixes the numerical Jacobian and also deals with SDC. We still need to sort out the interface to legacy Fortran networks, which will now integrate a different of variables.

zingale:
not that this numerical Jacobian is only used by Castro's true SDC at the moment.

zingale:
just as a check, we do still convert second-order with Strang with this change (as run with `reacting_convergence`:

```
 $\rho$                          2.794163e+18   2.044      6.777266e+17   2.554      1.154242e+17
 $\rho u$                        6.795828e+26   2.448      1.245025e+26   2.889      1.680538e+25
 $\rho v$                        6.795828e+26   2.448      1.245025e+26   2.889      1.680538e+25
 $\rho E$                        2.450825e+35   2.351      4.802854e+34   2.742      7.178876e+33
 $\rho e$                        2.260573e+35   2.320      4.526276e+34   2.821      6.403446e+33
 $T$                             2.237043e+21   1.691      6.926608e+20   2.482      1.239945e+20
 $\rho X(\isotm{He}{4})$         2.878085e+18   2.020      7.095873e+17   2.529      1.229473e+17
 $\rho X(\isotm{C}{12})$         1.697694e+17   1.950      4.392618e+16   2.232       9.35258e+15
 $\rho X(\isotm{O}{16})$         1.687142e+14   1.660      5.337997e+13   1.957      1.375142e+13
 $\rho X(\isotm{Fe}{56})$           279415400   2.044          67772660   2.554          11542420
```

maxpkatz:
To deal with Fortran, our wrapper is going to have to create temporary RHS and Jacobian arrays to pass to actual_rhs that have the temperature component, then strip the temperature component (for the RHS) or convert it to the energy component (for the analytic Jacobian).

maxpkatz:
This should be all set for final tests and review. The last change made was to set `do_constant_volume_burn` to true by default. The reasoning here is, now that we're explicitly integrating internal energy `e`, we should be consistent with the actual thermodynamic definition of `e` that pairs with the specific heat at constant volume. This makes the relationship with Maestro a little strained since that code integrates enthalpy, not energy. We will need to fix this in a later change, as this change is already large enough.

zingale:
in rprox and xrb_simple we override do_constant_volume_burn, so we should remove those overrides

zingale:
note: the Castro reacting_convergence test problem no longer works -- it gets CFL errors after 6 steps.  The problem seems to have entered between commits 697554d and cba9a1d

zingale:
final (?) Castro tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-02-15-003/index.html
don't worry about the SDC -- they are fixed in a separate PR

zingale (approved):

PR #499, opened 2021-01-01
define STRANG if we are not doing an SDC variant

zingale:
this will let us test on STRANG and simplify some preprocessor logic


maxpkatz (approved):

zingale:
I moved these into `Make.Microphysics_extern` so all codes benefit.

PR #503, opened 2021-01-02
Generalize linpack and move to util

### LINKED ISSUE #502: generalize vode_linpack.H
It would be nice if other integrators could use `vode_linpack.H`.  Right now, it depends on `vode_type.H` -- it looks like this is only for `VODE_NEQS`, so maybe we can just pass in the number of equations?  Then this can live in `integration/util/`

maxpkatz:
Closes #502


zingale (approved):

zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-01-01-003/index.html

PR #505, opened 2021-01-03
make the Microphysics parameters optionally managed by C++

zingale:
as part of this change, each parameter now has a namespace.  You can set them in your probin as before with no changes.  But you can also now set them in the inputs file, which means that they are read in by C++.  Here you would use the namespace for the parameter (like eos.use_coulomb = 1).  Note: if a parameter is set both in the probin and the inputs, then the inputs value is used.

This can be tested using the test_parameters unit test.


zingale:
all tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-01-09-003/index.html

maxpkatz (approved):

PR #506, opened 2021-01-03
Skip EOS reset if eos_t doesn't have relevant quantities

maxpkatz:
This is a first demonstration of how we can use C++17 `if constexpr` to template out unnecessary logic.


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-01-03/index.html

zingale (approved):

PR #507, opened 2021-01-03
Only access variables in the state in Helmholtz

maxpkatz:
Continues #470 by compiling out accesses to members of the struct sent to the EOS that are invalid. This will allow us later to, for example, send a type to the EOS that doesn't have entropy or enthalpy.

We don't explicitly compile out all intermediate expressions that led up to the quantities in question. That could be done, but compilers should be able to recognize that as dead code and compile it out for us.


zingale:
tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-01-03-001/index.html

zingale (approved):

PR #510, opened 2021-01-03
start reusing some code for runtime parameters

zingale:
This creates a centralized Params class in `runtime_paramerters.py` that eventually will be used by the application codes and Microphysics for managing runtime parameters.  It has all of the necessary methods to write the C++ and Fortran needed for when Fortran manages the runtime parameters (as currently is the case in Microphysics) and for when C++ manages them (like we do in Castro for the main parameters).


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-01-04-005/index.html
I think the errors are just because this adds `_rt` to the Real runtime parameters now, while before it was not.

util/build_scripts/write_probin.py
```diff
@@ -230,44 +202,37 @@ def write_probin(probin_template, param_files,
...
-
-                    else:
-                        fout.write("{}{}, allocatable, public :: {}\n".format(
-                            indent, p.get_f90_decl(), p.var))
+                    fout.write(f"{indent}{p.get_f90_decl_string()}")
 
                 if not params:
                     # we always make sure there is atleast one variable
-                    fout.write("{}integer, save, public :: a_dummy_var = 0\n".format(indent))
+                    fout.write(f"{indent}integer, save, public :: a_dummy_var = 0\n")
```
maxpkatz, line 209: Since you're using f-strings now, the shebang needs to be updated for python3

maxpkatz:
> 
> 
> tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-01-04-005/index.html
> I think the errors are just because this adds `_rt` to the Real runtime parameters now, while before it was not.

I agree that the old code is wrong, but it could also have been fixed before now by checking to replace `e` with `d`. I'm surprised we weren't already doing that, since I know it has come up before.

zingale:
Indeed, it turns out we weren't doing that replacement in Microphysics ("e" -> "d").

zingale:
updated to python3

maxpkatz (approved):

PR #513, opened 2021-01-04
Only access variables in the state in multigamma

maxpkatz:


zingale:
tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-01-04-002/index.html

zingale (approved):

PR #518, opened 2021-01-04
Add an EOS type that only has rho, T, e

maxpkatz:
Use it in the VODE Strang implementation as a proof of concept.


interfaces/eos_type.H
```diff
@@ -50,7 +50,29 @@ struct eos_t {
...
+    amrex::Real xn[NumSpec];
+#if NAUX_NET > 0
+    amrex::Real aux[NumAux];
+#endif
+    amrex::Real dedT;
+    amrex::Real dedr;
 
+    amrex::Real cv;
+    amrex::Real cp;
+
```
zingale, line 68: is it really possible to get cp when we don't have p?

maxpkatz: No

maxpkatz: If we want to merge this before we merge #496, we'll need to have a separate path for the case where we need cv and cp (which is straightforward).

zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-01-04-003/index.html

zingale:
the test that failed has call_eos_in_rhs = T, so I am not sure why this change affected that.

interfaces/eos_type.H
zingale: I'd be okay with merging #496 first.  I just want to study it a little more.

zingale:
once this is merged I think we can get rid of the `gamma_law` EOS in favor of just having `gamma_law_general`

maxpkatz:
Since we decided to block the burner change on #496, I'm removing the integration change for now; this PR just adds the new type. We can independently experiment with using it in other places like Castro hydro.

zingale (approved):

PR #535, opened 2021-01-15
change the if logic for the RHS include in C++ VODE

zingale:
in particular, this will allow Castro's true-SDC work


zingale:
all tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-01-16-003/index.html

maxpkatz (approved):

PR #536, opened 2021-01-16
Start moving towards zero-based indexing with linpack

maxpkatz:
`RArray1D` and `RArray2D` are one-based arrays, and that causes friction with the zero-based indexing of `xn` and other arrays. So we want to move everything to zero-based indexing.

The strategy will be to suffix function parameters with "1", e.g. `a` -> `a1`, and then do

```
auto& a = reinterpret_cast<Array2D<Real, 0, num_eqs-1, 0, num_eqs-1>&>(a1)
```

Then, in the function, we can convert all the calculations to zero-based indexing. When all of the conversions have been done, we can do one final sweep to convert `RArray1D` and `RArray2D` to be zero-based and then remove all the `reinterpret_cast` statements and rename the function parameters to their original names. This will allow us to do an incremental conversion.

### DIFF
--- a/util/linpack.H
+++ b/util/linpack.H
@@ -8,35 +8,38 @@
 
 template <int num_eqs>
 AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
-void dgesl (RArray2D& a, IArray1D& pivot, RArray1D& b)
+void dgesl (RArray2D& a1, IArray1D& pivot1, RArray1D& b1)
 {
+    auto const& a = reinterpret_cast<Array2D<Real, 0, num_eqs-1, 0, num_eqs-1>&>(a1);
+    auto const& pivot = reinterpret_cast<Array1D<int, 0, num_eqs-1>&>(pivot1);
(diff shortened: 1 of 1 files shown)


maxpkatz:
Credit to @WeiqunZhang for the idea for how to do this.

zingale:
all tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-01-16-001/index.html

util/linpack.H
```diff
@@ -64,19 +69,19 @@ void dgefa (RArray2D& a, IArray1D& pivot, int& info)
...
-            for (int i = k+1; i <= num_eqs; ++i) {
+            for (int i = k+1; i < num_eqs; ++i) {
                 if (std::abs(a(i,k)) > dmax) {
                     l = i;
                     dmax = std::abs(a(i,k));
                 }
             }
 
-            pivot(k) = l;
+            pivot(k) = l + 1;
```
zingale, line 85: why does pivot need to change here?

maxpkatz: `pivot` is not changing, `l` is (from one indexed to zero indexed)

zingale (approved):

PR #538, opened 2021-01-17
remove the Fortran VODE

zingale:
We no longer support the Fortran VODE since we did the port to C++ and all networks, including those with Fortran implementations, can be run from the C++ VODE.  Further the Fortran VODE no longer supports GPUs (since we switch to a C++ lambda-capture approach).


zingale:
as expected, all of the Fortran test_react tests fail:
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-01-17-001/index.html

maxpkatz (approved):

PR #541, opened 2021-01-20
Nuclear Partition Function Test Case

ChrisDeGrendele:
Don and I will go through this together soon. 

The main result:
```
Edot Ratio (PF/NOPF): 2.293683464                                                     
The estimated for 1e10k is 2.294                                                      
```


ChrisDeGrendele:
This pull request contains two main things. 

1. A single rate network.
2. `test_pf` This test is basically `test_rhs` run twice. Once with `use_partition_factors = .true.` and once with `use_partition_factors = .false.` We then construct the rhs and compare the two Edots. They should differ by the partition function factor of mg23 at the specific temperature. For now, I have hardcoded the solution into the code to compare to. The solution was generated using `test_pf/solution.ipynb`

zingale:
closing this -- we have a `test_part_func` unit test now.

PR #545, opened 2021-01-21
allow us to enable jacobian caching even on the GPU

### LINKED ISSUE #544: allow opt-in to Jacobian caching on GPUs in VODE
Currently, we disable Jacobian caching on GPUs with VODE, to save memory.  But for some nets, it seems to help a lot.  It would be nice to have a compile-time option to opt-in

zingale:
closes #544

### DIFF
--- a/integration/VODE/Make.package
+++ b/integration/VODE/Make.package
@@ -9,6 +9,15 @@ else
   endif
 endif
 
+# by default we do not enable Jacobian caching on GPUs to save memory
+ifneq ($(USE_CUDA), TRUE)
+  DEFINES += -DALLOW_JACOBIAN_CACHING
+endif
+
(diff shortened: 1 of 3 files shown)


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-01-21-003/index.html

dwillcox (approved):
Looks good!

dwillcox:
Looks like relative errors are O(1e-5) or less, except for Ni-56 in the rprox regtest, which at (abs=2e-5, rel=2e-3) is quite a bit larger than the species abs/rel tolerances of 1e-12 and 1e-8, respectively.

Maybe it's not really surprising that turning on Jac caching could cause that though, so I see no reason not to merge.

PR #547, opened 2021-01-23
iso7 17/∞

maxpkatz:
There are two parts to this change, which concentrate on the heavy-ion reactions (C12+C12, C12+O16, O16+O16).

1) Each of these is currently only written as a forward reaction, not a reverse reaction. However, the goal here is to refactor the rate infrastructure so that each rate automatically has forward and reverse components, rather than storing them as separate scalars as we do today. So we need to know what the reverse rate is, even though it's zero in this case. The reverse rates are added and just left as zero.

2) C12+O16 has two endpoints in this network, Mg24+He4 and Si28, and the two cases implicitly have equal probabilities. This is now made explicit by splitting this reaction into two rates, each of which is 0.5 * the old rate.

This change increases the number of rates in the network (although it still represents the same physics) and so slightly decreases the performance.

### DIFF
--- a/networks/iso7/actual_network.H
+++ b/networks/iso7/actual_network.H
@@ -47,9 +47,14 @@ namespace Rates {
                        C12_He4_to_O16_reverse,
                        He4_He4_He4_to_C12_forward,
                        He4_He4_He4_to_C12_reverse,
-                       C12_C12,
-                       C12_O16,
-                       O16_O16,
+                       C12_C12_to_Ne20_He4_forward,
+                       C12_C12_to_Ne20_He4_reverse,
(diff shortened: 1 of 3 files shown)


zingale (approved):

zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-01-23-001/test_react_C-iso7-VODE.html

zingale:
are you okay with that level of error?

maxpkatz:
> 
> 
> are you okay with that level of error?

Yes, I expect non-zero but small diffs since we're adding some terms and changing the order of operations on C12 + O16.

PR #548, opened 2021-01-23
Make Helmholtz data inline instead of extern

maxpkatz:


zingale (approved):

zingale:
all tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-01-23-002/index.html

PR #550, opened 2021-01-25
fix some Jacobian elements for eliminating T integration

zingale:


zingale:
with these changes, I think that Castro looks good.  I think this is ready to be merged into the original PR.

integration/utils/numerical_jacobian.H
```diff
@@ -104,14 +114,9 @@ void numerical_jac(burn_t& state, JacNetArray2D& jac)
         }
 
         for (int m = 1; m <= neqs; m++) {
-            jac(m, net_itemp) = 0.5_rt * (ydotp(m) - ydotm(m)) / (eps * state.T);
+            jac(m, net_ienuc) = 0.5_rt * (ydotp(m) - ydotm(m)) / (eps * state.e);
```
maxpkatz, line 117: This expression (which could be a divide-by-zero if `e` is zero) reminds me that in this new scheme, we really have to be careful that `state.e` is always the actual energy/enthalpy, not a counter starting from zero.

zingale:
we can add an asset checking e == 0

integration/VODE/vode_type_simplified_sdc.H
```diff
@@ -495,11 +495,11 @@ void jac_to_vode(const Real time, burn_t& state,
 
     // d( d(rho X_m)/dt)/dT
     for (int m = 1; m <= NumSpec; m++) {
-        dRdw(SFS+m, iwT) = state.rho * jac_react(m, net_itemp);
+        dRdw(SFS+m, iwT) = state.rho * jac_react(m, net_ienuc);
```
maxpkatz, line 498: We should change this variable name `iwT` and the comment above

maxpkatz:
We should probably update the definition of `w` in the comments explaining the `dRdw` / `dwdU` scheme

maxpkatz (approved):

integration/VODE/vode_type_simplified_sdc.H
zingale: done

PR #553, opened 2021-01-30
iso7 18/∞

maxpkatz:
Three reactions in iso7 include reactions of identical particles. At the moment, the factor of 1/2 implied by this is not included in the reaction rate, but rather applied in the RHS directly.

To continue the RHS generalization process we're doing, it will be useful if we don't have to worry about this factor, so this PR moves the term directly into the reaction rate, multiplying the RHS by the inverse factor.

For example, the C12 + C12 term currently looks like

```
dydt(He4)  = (1/2) * Y(C12) * Y(C12) * rate(ir1212)
dydt(C12)  =        -Y(C12) * Y(C12) * rate(ir1212)
dydt(Ne20) = (1/2) * Y(C12) * Y(C12) * rate(ir1212)
```

and it will now look like

```
rate(ir1212) *= 1/2;

dydt(He4)  =      Y(C12) * Y(C12) * rate(ir1212)
dydt(C12)  = -2 * Y(C12) * Y(C12) * rate(ir1212)
dydt(Ne20) =      Y(C12) * Y(C12) * rate(ir1212)
```

This makes it very clear that the reaction produces one He4 nucleus and one Ne20 nucleus, and consumes two C12 nuclei. This rewrite is practically useful (see below), and if we choose to think of it this way, it's generally fair to consider the identical particle factor as a modification of the cross-section, instead of a term that appears explicitly in the RHS.

The step that this helps enables next is to generalize the RHS term further. We want to write that the forward rate `ir1212` is a special case of the reaction class A + B -> C + D, which is one of the classes we can apply automated reaction infrastructure to (see e.g. the appendix of [this paper](https://iopscience.iop.org/article/10.1088/1749-4699/6/1/015003/pdf)). In our implementation of the automated infrastructure, all we need to do is provide the species indices of A, B, C, and D, as well as the reaction rate, and the infrastructure then adds the relevant contributions to the RHS for each species. The factor of 2 would automatically be captured as a result of A and B being the same.

### DIFF
--- a/networks/iso7/actual_network.H
+++ b/networks/iso7/actual_network.H
@@ -95,7 +95,7 @@ namespace RHS {
             break;
 
         case rhs_rate<NumRates>(He4, He4_He4_He4_to_C12_forward):
-            data.prefactor  = -1.0_rt;
+            data.prefactor  = -3.0_rt;
             data.specindex1 = He4;
             data.specindex2 = He4;
             data.specindex3 = He4;
@@ -174,7 +174,7 @@ namespace RHS {
             break;
 
         case rhs_rate<NumRates>(C12, He4_He4_He4_to_C12_forward):
-            data.prefactor  = 1.0_rt / 3.0_rt;
+            data.prefactor  = 1.0_rt;
             data.specindex1 = He4;
             data.specindex2 = He4;
             data.specindex3 = He4;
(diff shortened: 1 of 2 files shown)


maxpkatz:
Also, a note for posterity because it temporarily confused me: the triple alpha reaction has an identical particle factor of 1/2, not 1/6, because it's really He4 + He4 -> Be8 and then Be8 + He4 -> C12, so the cross-section factor only applies on the first step in the chain.

zingale:
I don't think that's right.  I think it has a -3 because three particles are consumed and then a 1/6 for the particle factor.

maxpkatz:
Oh, you're of course right. Updated.

zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-01-30-002/test_react_C-iso7-VODE.html

maxpkatz:
The standard level of small but non-zero diff is expected here due to change in order of float operations.

zingale (approved):

PR #554, opened 2021-01-31
iso7 19/∞

maxpkatz:
Move y(Si28) into the RHS instead of the rate. This will be important later for calculating the equilibrium state.


maxpkatz:
I'm reversing the order of the changes, pushing off the larger refactor for the moment, so now this PR is just a simple change that moves the silicon abundance out of the silicon -> nickel rate and directly into the RHS.

zingale (approved):

PR #555, opened 2021-02-01
fix create_release action

zingale:
we were not sending the release text for the environment


zingale:
note: I want this on `main` so it takes effect next release

maxpkatz (approved):

PR #558, opened 2021-02-03
extract out the Jacobian into a normal .cpp

zingale:
The CUDA build with simplified SDC was taking a long time (> 2 hours).  By extracting the Jacobian bits into a .cpp file instead of inlining them, the build time is greatly reduced.  This addresses Castro issue: https://github.com/AMReX-Astro/Castro/issues/1567


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-02-03-001/index.html
the one failure is just an old benchmark

harpolea (approved):

PR #560, opened 2021-02-03
Get simplified SDC to work with MAESTROeX

doreenfan:
This PR fixes some bugs seen with simplified SDC and SDC_EVOLVE_ENTHALPY on.


zingale:
we haven't previously relied on accessing a code's runtime parameters directly in Microphysics.  I wonder if we should add the use_tfromp to the burn_t instead?

maxpkatz:
Long term, yes we should do that. Short term, I'm fine with this, because we're going to re-architect things anyway so that the burner does the right thing when enthalpy is the input versus when energy is the input. We can handle tfromp at that time.

zingale (approved):

PR #563, opened 2021-02-06
iso7 20/∞

maxpkatz:
The general form of a reaction is
```
n_a A + n_b B + n_c C <-> n_d D + n_e E + n_f F
```
for species A, B, C, D, E, and F, where n_a particles of A, n_b particles of B, and n_C particles of C are consumed in the forward reaction and produced in the reverse reaction, and the analogous statement holds for particles D, E, and F.

For a given species, such as species A, the forward reaction term is of the form
```
-n_A * Y(A)**a * Y(B)**b * Y(C)**c * forward_rate
```
and the reverse reaction term is of the form
```
n_A * Y(D)**d * Y(E)**e * Y(F)**f * reverse_rate
```
Here a, b, and c are reaction-specific exponents which usually, but not always, are equal to n_a, n_b, and n_c respectively.

For example, in C12 + He4 <-> O16, species A is C12, species B is He4, species D is O16 (the other species are unused). Then n_a = n_b = n_d = 1, and a = b = d = 1. In the triple alpha forward reaction we have A = He4, D = C12, n_a = 3, a = 3, and n_d = 1.

In order to do this, we add a `rate_fr_t` which explicitly holds both forward and reverse contributions, and we take the shortcut (for now) that it should have half as many rates as the "real" network rates (which separate out the forward and reverse terms). After the Jacobian is converted over to this new scheme, we can make things a little clearer.

By writing it in this way, it becomes possible to explicitly identify reaction pairs that are in equilibrium and then take some action in response (this will be the focus of a later PR in this series).


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-03-02-002/index.html

zingale:
the structural changes look good here.  The test_react test is very different, and takes a different number of RHS calls, but I think that is just because a few zones are bouncing around differently in equilibrium.

zingale (approved):

zingale:
The general form of a reaction is

n_a A + n_b B + n_c C <-> n_d D + n_e E + n_f F

for species A, B, C, D, E, and F, where n_a particles of A, n_b particles of B, and n_C particles of C are consumed in the forward reaction and produced in the reverse reaction, and the analogous statement holds for particles D, E, and F.

For a given species, such as species A, the forward reaction term is of the form

-n_A * Y(A)**a * Y(B)**b * Y(C)**c * forward_rate

and the reverse reaction term is of the form

n_A * Y(D)**d * Y(E)**e * Y(F)**f * reverse_rate

Here a, b, and c are reaction-specific exponents which usually, but not always, are equal to n_a, n_b, and n_c respectively.

For example, in C12 + He4 <-> O16, species A is C12, species B is He4, species D is O16 (the other species are unused). Then n_a = n_b = n_d = 1, and a = b = d = 1. In the triple alpha forward reaction we have A = He4, D = C12, n_a = 3, a = 3, and n_d = 1.

In order to do this, we add a rate_fr_t which explicitly holds both forward and reverse contributions, and we take the shortcut (for now) that it should have half as many rates as the "real" network rates (which separate out the forward and reverse terms). After the Jacobian is converted over to this new scheme, we can make things a little clearer.

By writing it in this way, it becomes possible to explicitly identify reaction pairs that are in equilibrium and then take some action in response (this will be the focus of a later PR in this series).

PR #565, opened 2021-02-10
add checks on O and Si to the NSE check

zingale:
this allows us to better capture the regions where NSE should hold


zingale:
all tests pass:

http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-02-10-003/index.html

harpolea (approved):

PR #566, opened 2021-02-11
aprox13 2/∞

maxpkatz:
Move identical particle factor into C12+C12 rate. Also add an entry for the reverse rate which will be zero always.


zingale (approved):

zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-02-11-001/index.html

PR #567, opened 2021-02-11
Remove Jacobian terms that depend on d(rate)/dY

maxpkatz:
From experimentation, these don't seem to be especially helpful for most networks, and they will get in the way of the automated infrastructure for building a Jacobian from the RHS. We may be able to add them back later after all of the networks have been refactored, if needed.


zingale:
here are the Castro tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-02-11-003/index.html

maxpkatz:
Evidently that shows how much of an effect a different Jacobian results in our answer. (Which admittedly just reveals that rtol_spec doesn't mean a whole lot.) I measure the effect of this change by the question of whether it results in a comparable number of average RHS calls, which it does. For aprox19 the maximum goes from 3065 to 3261, and for iso7 it goes from 17408 to 4654, but I think we've come to understand that this maximum just measures whether we got lucky or not when we arrived at NSE.

zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-06-08-006/index.html

zingale (approved):

zingale:
test diffs are small: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-06-08-007/index.html

zingale:
note: some tests with 0 diffs use the numerical Jacobian.

Issue #568, opened 2021-02-14
Screening factors should not depend on `jscr`


maxpkatz:
For the aprox nets, we currently construct an array of screening factors at setup time and then index into them with an integer during rate calculation.

In the future, we would like to have the ability to construct the rates one by one, rather than in a large loop (so that we could experiment with removing the rate_t and constructing RHS terms on demand). This means that we cannot maintain the assumption of a specific order in which the screening terms are evaluated.

The current system is generally fragile anyway, so this also presents an opportunity for a general code cleanup.

maxpkatz:
Support for this was added in #726. At this time it has only been done for aprox13 in #727, but it will be done for the other networks soon, so this can be closed.

PR #569, opened 2021-02-14
aprox13 3/∞

maxpkatz:
This change folds the reaction pairs (irmgap, irr1), which did Mg24 -> Al27 -> Si28, and (irsigp, irr1), which did Si28 -> Al27 -> Mg24, into the effective Mg24 <-> Si28 rates. This allows us to remove irr1 from the rate list. More importantly it works toward the goal where every "reaction" known to the network connects an actual pair of species in the network, while any intermediate rates connecting the two can be considered implementation details, which will be helpful for methods looking at microscopic equilibria. This change will be followed by removing the rest of the intermediate proton links.

### DIFF
--- a/networks/aprox13/actual_network.H
+++ b/networks/aprox13/actual_network.H
@@ -102,7 +102,6 @@ namespace Rates {
                         ircopa,
                         ircopg,
                         irnigp,
-                        irr1,
                         irs1,
                         irt1,
                         iru1,
--- a/networks/aprox13/actual_network_data.cpp
+++ b/networks/aprox13/actual_network_data.cpp
@@ -102,7 +102,6 @@ void actual_network_init()
         names[ircopa-1] = "rcopa";
         names[ircopg-1] = "rcopg";
         names[irnigp-1] = "rnigp";
-        names[irr1-1]   = "r1";
         names[irs1-1]   = "s1";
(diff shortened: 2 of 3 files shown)


maxpkatz:
Note: irmgap, the rate for Mg24(a,p)Al27, is still in the list of rates known to the network. A requirement of the target scheme is that this no longer be true, and it would just be calculated inline while calculating the effective Mg24(a,g)Si28 rate. However it is not straightforward to do this yet -- at the moment, we calculate at once all of the rates in the `aprox13rat` or `aprox13` tab routine, and we don't have a way to select a specific rate for calculation, which is what would be needed here. So a later change will be to convert the rate routines to calculate rates one-by-one (see also #568), and then we can do that second piece of cleanup.

zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-02-14/index.html

in general these look fine, but do we understand why the Jacobian terms differ so much in test_rhs?

maxpkatz:
I had a sign error in one of the d/dT terms. Fixed.

zingale:
updated tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-02-14-001/index.html

zingale (approved):

PR #570, opened 2021-02-14
aprox13 4/∞

maxpkatz:
Calculate density dependence in tabulated rates inline. This will help in a future step where we can query `aprox13tab` a rate at a time instead of computing all the (unscreened) rates at once.


zingale:
all tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-02-14-002/index.html

zingale (approved):

PR #571, opened 2021-02-14
iso7 21/∞

maxpkatz:
Calculate density dependence in tabulated rates inline. This will help in a future step where we can query `iso7tab` a rate at a time instead of computing all the (unscreened) rates at once.


networks/iso7/actual_rhs.H
```diff
@@ -108,17 +81,113 @@ void iso7tab(const Real btemp, const Real bden,
...
+        case (C12_O16_to_Mg24_He4_reverse):
+            dtab = bden; // rate is zero in this net
+            break;
+
+        case (C12_O16_to_Si28_forward):
+            dtab = bden;
+            break;
+
+        case (C12_O16_to_Si28_reverse):
+            dtab = bden; // rate is zero in this net
```
zingale, line 129: this should be 1

maxpkatz: fixed

zingale (approved):

Issue #20, opened 2016-11-08
need to reevaluate the tolerances


zingale:
We ask for species to be evolved to a tolerance of 1.d-12 (in `integration/_parameters`).

This is pretty tight.  We need to check whether it can be relaxed.  We can relax on a network-by-network basis (using priorities in the `_parameter` files).  

It seems that the original aprox13 and aprox19 networks used tolerances of 1.e-6

zingale:
note also that the `BS` integrator has only a single tolerance -- it takes the max of all the tolerances:

```
    bs % atol(:) = 0.d0                                                         
    bs % rtol(:) = max(rtol_spec, rtol_temp, rtol_enuc)                         
```

zingale:
the way to properly do this is to look at convergence of the results as we tighten the tolerances for each burner / integrator.

zingale:
there is some thought on this in the wdmerger II paper.  It seems that 1.e-8 should be okay...

zingale:
I think that @abigailbishop largely has done this.  We should add some plots to the user's guide

abigailbishop:
I can do that.

zingale:
I think we've mostly addressed this now.

PR #582, opened 2021-02-18
aprox19 1/∞

maxpkatz:
Move identical particle factors into rates

### DIFF
--- a/networks/aprox19/actual_rhs.H
+++ b/networks/aprox19/actual_rhs.H
@@ -308,6 +308,10 @@ void screen_aprox19 (Real btemp, Real bden,
     dratdumdt(irg3a) = dratrawdt(irg3a)*sc3a + ratraw(irg3a)*sc3adt;
     //dratdumdd(irg3a) = dratrawdd(irg3a)*sc3a + ratraw(irg3a)*sc3add;
 
+    // Identical particle factor (He4 + He4 + He4)
+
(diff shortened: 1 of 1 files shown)


maxpkatz:
Note to self for later: the number of particles consumed in the proton-proton reaction is 3 even though the term goes as Y(H1)**2. This will require the same kind of special handling as the silicon burning reaction in iso7 where 7 alphas are consumed even though the term goes as Y(He4).

zingale (approved):

zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-02-18-004/index.html

zingale:
I think those errors look reasonable

zingale:
why does the pp consume 3?  is it merging p+p and the d+p?

maxpkatz:
Yes, that's right.

I really want it to be the case that we define an RHS term as -N * Y**N, so that the number of particles you consume is exactly the same as the exponent on Y in the term. Almost every term meets that requirement trivially with N == 1, but then there are exceptions like this one that do not. So I think what I will do (assuming no objections) is to make this the case and then multiply or divide the rate by the corresponding factors of Y to offset this. It will make the refactor in iso7 #20 even simpler.

zingale:
or just add deuterium?

maxpkatz:
Yes, that's also a solution, although we should make a call about how much we care about the goal of having these networks be notionally the same as the other instantiations in the literature.

PR #585, opened 2021-02-18
Remove network_rhs.F90 which is no longer used

maxpkatz:


zingale (approved):

zingale:
all tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-02-18-002/index.html

PR #586, opened 2021-02-18
Remove unused rate names

maxpkatz:
The string names of the rates aren't actually used anywhere, and since we're in the middle of refactors involving lots of rate removals/additions and name changes, these will just get in the way.


zingale (approved):

zingale:
all tests pass:
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-02-18-003/index.html

PR #590, opened 2021-02-21
Make network and RHS data inline

maxpkatz:


zingale:
aprox19 + NSE fails to compile:
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-02-21/index.html

maxpkatz:
Closing this, I need to redo it with a better approach later

PR #591, opened 2021-02-21
Remove constant pressure burn code path in C++

maxpkatz:
This will need to be replaced later with a correct solution for integrating enthalpy


zingale:
all tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-02-21-001/index.html

zingale (approved):

PR #597, opened 2021-02-23
Re-enable call_eos_in_rhs option

maxpkatz:
For now, this is done for Strang only.

Note: in principle this could break a setup that was previously explicitly setting `call_eos_in_rhs = F`. However, I think this is pretty unlikely to have happened, I only found one setup (Castro xrb_mixed) that is doing it.

### DIFF
--- a/integration/ForwardEuler/actual_integrator.H
+++ b/integration/ForwardEuler/actual_integrator.H
@@ -68,7 +68,9 @@ void clean_state (burn_t& state)
 
     // Evaluate the EOS to get T from e.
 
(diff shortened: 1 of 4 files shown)


zingale (approved):

zingale:
NSE test failed for some reason:
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-02-25-002/index.html

zingale:
Oh, NSE failed because it was explicitly setting `call_eos_in_rhs=0`

PR #617, opened 2021-03-07
iso7 22/∞

maxpkatz:
For all Jacobian terms of the form d(f) / d(Y), we can compute these automatically based on RHS, since we know how to compute the derivative of a term like (N * Y_1**e_1 * Y_2**e_2 * rate) analytically (multiply by the exponent, then decrement the exponent). This means we do not need to explicitly hand-write these terms to construct the network.


zingale:
tests look reasonable:
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-03-07-001/index.html

zingale (approved):

PR #618, opened 2021-03-07
Move legacy constants into fundamental_constants header

maxpkatz:


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-03-07-002/index.html

maxpkatz:
rprox compilation should be fixed. This should change rprox results a little though since the value of the constants is slightly different from what the aprox networks had.

zingale:
tests look reasonable: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-03-10-001/index.html

zingale (approved):

PR #619, opened 2021-03-07
add VODE step rejection for true SDC based on rho X

zingale:


zingale:
Castro tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-03-16-002/index.html

integration/VODE/vode_type_simplified_sdc.H
```diff
@@ -127,6 +127,35 @@ void renormalize_species(const Real time, burn_t& state, dvode_t& vode_state)
...
+void renormalize_species(burn_t& state)
+{
+
+    // Constrain the partial densities in burn_t state to sum to the 
+    // density.
+    //
+    // This is meant to be used upon exit, and we assume that
+    // vode_to_burn was already called
+
+#ifdef SDC_EVOLVE_ENERGY
```
maxpkatz, line 140: Is there a reason we cannot do this for the enthalpy burn?

zingale: we don't have SRHO in the state, for some reason, so it needs to be done differently, and at the moment, I don't have anything to test it with.

maxpkatz (approved):

PR #620, opened 2021-03-08
Do normal sum in the new RHS scheme

maxpkatz:


zingale:
Castro tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-03-07-006/index.html

zingale:
Microphysics tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-03-08/index.html

zingale:
aprox13 takes about the same number of steps with and without this change, but iso7 seems to need more steps (for the worst case) with this change.

maxpkatz:
That metric can be misleading. I saw exactly the opposite effect running test_react on a different system. Since it's mostly related to order of operations working out well (or not), it can even be different on which compiler you're using.

zingale (approved):

PR #622, opened 2021-03-09
iso7 23/∞

maxpkatz:
Use normal sum in iso7 Jacobian


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-03-09-002/index.html

zingale (approved):

PR #624, opened 2021-03-13
iso7 24/∞

maxpkatz:
Move generic species Jacobian into utilities header


zingale (approved):

zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-03-13/index.html
do we expect this to change answers to roundoff?

maxpkatz:
Yes. Previously we were adding the d(rate)/dY terms in the same loop as these primary terms, now we're splitting them into separate loops.

PR #625, opened 2021-03-13
iso7 25/∞

maxpkatz:
Convert tabulated rates to new rate indexing scheme


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-03-13-002/index.html

zingale:
although I am not sure we are testing tabulated rates...

zingale:
this fails: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-07-19-002/test_react_C-iso7-VODE-table.html

maxpkatz:
fixed

zingale:
updated test: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-07-19-003/index.html

zingale:
Note: I had the test name setup wrong which is why the non-table version crashed.  But conveniently, this shows us the error from tables vs no tables :)

networks/iso7/actual_rhs.H
```diff
@@ -81,113 +81,91 @@ void iso7tab(const Real btemp, const Real bden,
...
-            dtab = 1.0_rt; // rate is zero in this net
+        case (Ca40_He4_to_Ti44):
+            forward_dtab = bden;
+            reverse_dtab = 1.0e0_rt;
             break;
 
-        case (O16_O16_to_Si28_He4_forward):
-            dtab = bden;
+        case (Si28_7He4_to_Ni56):
+            forward_dtab = 0.0e0_rt;
```
zingale, line 144: why is this 0?

networks/iso7/actual_rhs.H
```diff
@@ -81,113 +81,91 @@ void iso7tab(const Real btemp, const Real bden,
...
-
-        case (Ca40_He4_to_Ti44_reverse):
-            dtab = 1.0e0_rt;
-            break;
+        }
 
-        case (Si28_7He4_to_Ni56_forward):
-            dtab = 0.0e0_rt;
-            break;
+        rate(2 * rateindex - 1) = (alfa * rattab(rateindex, 1, iat  ) +
```
zingale, line 150: can you add a comment explaining what this indexing means?

zingale: also, why not just fill `rattab` directly here?

networks/iso7/actual_rhs.H
maxpkatz: The Si <-> Ni rate in this network isn't evaluated like a traditional rate; we calculate the Ca <-> Ti rate and then in the screening routine postprocess that into an effective Si <-> Ni rate (even though it's not really related to screening). So there's nothing to do yet by this point, everything in forming this rate happens later.

zingale: okay just checking

maxpkatz: I'll add a comment so we don't think that's a mistake later.

networks/iso7/actual_rhs.H
maxpkatz: I am not sure I understand the question, rattab is the table and is not modified in this routine.

maxpkatz: I added a comment as requested, but note the legacy indexing is going to go away soon, and every rate will have both a forward and a backward component rather than being independent entries in a flattened array, so the current setup is just temporary.

zingale (approved):

PR #627, opened 2021-03-14
update nova with pynucastro so it builds again

zingale:


networks/nova/actual_network.F90
```diff
@@ -78,7 +78,7 @@ module actual_network
 
   real(rt), allocatable, save :: bion(:), mion(:)
 
-#if defined(AMREX_USE_CUDA) && defined(AMREX_USE_GPU_PRAGMA)
+#ifdef AMREX_USE_CUDA
```
maxpkatz, line 81: This won't work with CUDA, but none of the Fortran networks will at this point anyway.

zingale: indeed, I am not worrying about CUDA yet.  At some point we'll just switch things over to C++

maxpkatz (approved):

PR #637, opened 2021-03-26
enable the constraint on how much the mass fractions can change per step

zingale:
this is now on with simplified-SDC


maxpkatz (approved):

zingale:
now that we have the rtol fix in place (#643), this seems like it is too restrictive and causes problems.

PR #640, opened 2021-04-05
cap the timestep so we don't go beyond tout

zingale:
this attempts to do the logic that was in the original VODE (itask = 4) to prevent the timestep from taking us beyond tout.


zingale:
The main issue was fixed by #643 
We might want to revisit this in the future, but I don't think this implementation is correct.

PR #641, opened 2021-04-05
enforce that species can't change too much per step

zingale:
this is for simplified SDC + VODE


zingale:
this is the same as #637

PR #643, opened 2021-04-05
fix scaling of rtol for species in SDC

zingale:
it should be left unscaled.  only atol gets scaled by rho


zingale:
this appears to fix:

https://github.com/AMReX-Astro/Castro/issues/1596
https://github.com/AMReX-Astro/Castro/issues/1597

maxpkatz (approved):

PR #645, opened 2021-04-06
add some more checks for true SDC

zingale:


zingale:
we've eliminated the separate code path for true and simplified SDC in the integrator, so this no longer applies

PR #646, opened 2021-04-06
convert the CJ detonation code to C++

zingale:


zingale:
Here's the output:

cj_det

maxpkatz (approved):

PR #647, opened 2021-04-12
reject a step if sum{rhoX} != rho to tolerance

zingale:
this applies only for simplified SDC


zingale:
this is no longer needed with #815

PR #650, opened 2021-04-14
add 2 new EOS types

zingale:
eos_rep_t has everything needed for p and e


zingale:
this compiles in Castro if I use the new type in the BCs

maxpkatz (approved):

PR #651, opened 2021-04-14
remove dpde and dpdr_e -- these were only used once in Castro

zingale:
we can compute them as needed on the fly

### DIFF
--- a/EOS/breakout/actual_eos.F90
+++ b/EOS/breakout/actual_eos.F90
@@ -123,9 +123,6 @@ contains
        ! sound speed
        state % cs = sqrt(gamma_const * poverrho)
 
-       state % dpdr_e = poverrho
-       state % dpde = (gamma_const-ONE) * state % rho
-
        ! Try to avoid the expensive log function.  Since we don't need entropy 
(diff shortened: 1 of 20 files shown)


zingale:
this depends on https://github.com/AMReX-Astro/Castro/pull/1625

zingale:
there are 2 problems with this:

breakout doesn't seem to define the needed derivatives to compute this as needed
polytrope has dedT = 0, which is in the denominator of the quantity we need to compute

maxpkatz:
We should require that every network at least fills out all of the derivatives even if they're zero. We can detect that in Castro and set dpde and dpdr_e to zero accordingly.

zingale:
agreed.  That was my plan.  I just haven't gotten back to this.

maxpkatz:
As I was working on the Tillotson EOS I realized my comment above isn't really right. For that EOS the natural variables are (rho, e) and T isn't actually needed to evaluate the EOS. So while dP/de is a natural variable, dP/dT and de/dT are not (although an arbitrary T can be constructed). Since this is the same problem polytrope has, maybe we should just not worry about this change.

zingale:
I agree, we can drop this change

PR #653, opened 2021-04-14
remove the gamma_law EOS and rename gamma_law_general to gamma_law

zingale:
we no longer need the simpler version of this EOS since we have templated EOSes
and can define types that carry a reduced set of information.


maxpkatz (approved):

zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-04-14-001/index.html

PR #660, opened 2021-04-30
make the numerical Jacobian more robust

zingale:
we now use the algorithm from LSODE / VODE.  In particular, you can find the description of this implementation in Radhakrishnan and Hindmarsh, "Description and Use of LSODE, the Livermore Solver for Ordinary Differential Equations", section 3.4.5


zingale:
we don't seem to have test coverage for this:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-07-08-004/index.html

zingale:
this seems to work for the SDC massive_star problem

zingale:
Here's the difference with this Jacobian change:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-07-12-002/massive_star-sdc.html

integration/utils/numerical_jacobian.H
```diff
@@ -15,15 +15,19 @@
...
+struct jac_info_t {
+    Real h;
+};
 
+const amrex::Real U = std::numeric_limits<amrex::Real>::epsilon();
 
-    // the choice of eps should be ~ sqrt(eps), where eps is machine epsilon.
-    // this balances truncation vs. roundoff error in the differencing
+AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+void numerical_jac(burn_t& state, const jac_info_t jac_info, JacNetArray2D& jac)
```
maxpkatz, line 28: We should pass the jac_info by reference (it could always get bigger later).

zingale: done

maxpkatz (approved):

PR #665, opened 2021-05-03
we no longer use rtol_temp and atol_temp, so remove them

zingale:
note: any application probin's that use these need to be updated


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-05-03-002/index.html

maxpkatz (approved):

PR #666, opened 2021-05-03
remove the Fortran numerical Jacobian -- it is not used

zingale:


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-05-03-003/index.html

maxpkatz:
Can you confirm this is no longer used in either Castro or Maestro?

zingale:
neither code uses it

maxpkatz (approved):

PR #674, opened 2021-05-25
don't try to read probin if it doesn't exist

zingale:


interfaces/extern_probin.template
```diff
@@ -55,28 +55,31 @@ subroutine runtime_init(name,namlen)
...
-  if (status < 0) then
-     ! the namelist does not exist, so we just go with the defaults
-     continue
+     if (status < 0) then
+        ! the namelist does not exist, so we just go with the defaults
+        continue
 
-  else if (status > 0) then
-     ! some problem in the namelist
-#ifndef AMREX_USE_CUDA
```
maxpkatz, line 73: This ifdef is unnecessary in this code anyway

zingale: removed

maxpkatz (approved):

PR #676, opened 2021-05-29
Mark some variables maybe_unused in Helmholtz

maxpkatz:
This attribute (from C++17) quiets unused variable compiler warnings.


zingale:
maybe we should replace ``MICROPHYSICS_UNUSED`` then?

zingale:
did we decide if we were going to do this via MICROPHYSICS_UNUSED? or should I just merge?

maxpkatz:
Yes, I agree that we should do it via MICROPHYSICS_UNUSED, so please hold for now until I make that change.

maxpkatz:
I changed my mind on the above, we should try to use the C++ standard syntax, and only drop back to using the macro if we run into a compiler that doesn't support this. So we should merge. I'll deal with MICROPHYSICS_UNUSED some other time.

zingale (approved):

zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2022-04-30/index.html

PR #677, opened 2021-05-30
for C++, we need to change logical to integer for the job_info test

zingale:


maxpkatz (approved):

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-05-30-010/index.html

PR #680, opened 2021-06-07
allow for runtime parameter arrays to be read at runtime

zingale:
this adds the C++ queryarr logic to support specifying values for array
runtime parameters in an inputs file or on the commandline

### DIFF
--- a/util/build_scripts/runtime_parameters.py
+++ b/util/build_scripts/runtime_parameters.py
@@ -184,7 +184,18 @@ class Param:
 
         ostr = ""
         if language == "C++":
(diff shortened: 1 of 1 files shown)


zingale:
tested this with problem.center in Castro.  This generates the code:

```
        for (int n = 0; n < 3; n++) {
            problem::center[n] = 0.0e0_rt;
        }
        
        amrex::Vector<amrex::Real> center_tmp(3, 0.0e0_rt);
        if (pp.queryarr("center", center_tmp, 0, 3)) {
            for (int n = 0; n < 3; n++) {
                problem::center[n] = center_tmp[n];
            }

        }
```

maxpkatz (approved):

PR #681, opened 2021-06-07
move the check that the temperature is reasonable after we update thermo

zingale:
previously, we were checking on the old T, not the T corresponding to the
current e that we are trying to integrate with.

### DIFF
--- a/integration/VODE/vode_rhs.H
+++ b/integration/VODE/vode_rhs.H
@@ -25,6 +25,14 @@ void rhs (const Real /*time*/, burn_t& state, dvode_t& vode_state, RArray1D& ydo
     // y(1:NumSpec) = dX/dt
     // y(net_ienuc) = denuc/dt
 
+    // Fix the state as necessary.    
+
+    clean_state(vode_state);
+
+    // Update the thermodynamics as necessary.
+
+    update_thermodynamics(state, vode_state);
+
     // Only do the burn if the incoming temperature is within the temperature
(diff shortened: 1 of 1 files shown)


zingale:
this "fixes" the FPE we see in the test suite for the massive star problem with NSE.

But I am not 100% certain that our VODE logic will reject the step in this case.

zingale:
actually, since we have the burn_state in vode_dvstep.H, I think I should put a step rejection if T > MAX_TEMP.  Then I think this will pick up on that.

maxpkatz:
Yes I agree that there is potentially still an issue to be solved in making sure the step gets rejected, however this is a clearly correct change when viewed in isolation so we should still do this.

zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-06-07-002/index.html

maxpkatz (approved):

maxpkatz:
The test failures are because some of these tests do hit T > 1e11. So while this is fine, we may want to take some other action if we detect that we have hit MAX_TEMP. This temperature is high enough that we don't expect it to be hit during a normal burn.

zingale:
interestingly, for  test_react_C-aprox19-VODE, we trigger the temperature check 2 times with this change, and in both cases, it is because the temperature is too cold, not to hot.

zingale:
with iso7, we have a few instances of temperature floor and about 2 dozen T > MAX_TEMP cases.

PR #682, opened 2021-06-07
add some additional VODE step rejection checks for Strang

zingale:


zingale:
curiously this has no effect: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-06-08-005/index.html

PR #683, opened 2021-06-08
remove some unnecessary calls in Strang

zingale:
update_thermodynamics already ensures that the burn_t is valid, so we don't need a vode_to_burn (or later burn_to_vode) call in vode_rhs.


zingale:
all tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-06-08-003/index.html

maxpkatz (approved):

PR #686, opened 2021-06-09
Remove Jacobian infrastructure for d(rate)/dY

maxpkatz:
Now that #567 is done, we can remove the custom infrastructure for the Jacobian. Now in iso7, both the RHS and the Jacobian are automatically constructed from a single source of truth in actual_network.H.

### DIFF
--- a/interfaces/network_utilities.H
+++ b/interfaces/network_utilities.H
@@ -26,18 +26,6 @@ struct rhs_t {
     int exponent_F;
 };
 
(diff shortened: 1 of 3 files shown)


zingale:
all tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-06-10-001/index.html

zingale (approved):

PR #689, opened 2021-06-10
create USE_FORT_MICROPHYSICS to enable/disable the Fortran network build

zingale:
this works in tandem with USE_FORT_EOS

### DIFF
--- a/networks/Make.package
+++ b/networks/Make.package
@@ -106,5 +106,3 @@ $(NETWORK_OUTPUT_PATH)/network_properties.F90:
            --defines "$(DEFINES)"
 
 endif
-
-test_network_header: $(NETWORK_OUTPUT_PATH)/network_properties.H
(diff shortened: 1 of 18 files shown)


zingale:
all tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-06-10-002/index.html

networks/Make.package
```diff
@@ -100,4 +108,3 @@ $(NETWORK_OUTPUT_PATH)/network_properties.F90:
 endif
 
 test_network_header: $(NETWORK_OUTPUT_PATH)/network_properties.H
```
maxpkatz, line 110: I guess we don't need this one anymore

zingale: eh, I like having those debug targets in the code just so I can quickly test changes

maxpkatz: right but it now does exactly the same thing as net_prop_debug

zingale: ah indeed.  I'll remove it.

zingale: done

maxpkatz (approved):

PR #693, opened 2021-06-12
iso7 26/∞

maxpkatz:
Store unused reverse rates in the rates array -- this will end up happening automatically in a subsequent change anyway. This change has no effect since these rates are set to zero in the rate routines.


zingale:
all tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-07-22-001/index.html

zingale (approved):

PR #694, opened 2021-06-19
have runtime_init bind to C so we can call it directly

zingale:


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-06-19-001/index.html

maxpkatz (approved):

PR #695, opened 2021-06-20
since we've merged the c12ag rates, modify the test_aprox_rates function calls

zingale:
this also now builds C++ only

### DIFF
--- a/unit_test/test_aprox_rates/GNUmakefile
+++ b/unit_test/test_aprox_rates/GNUmakefile
@@ -18,6 +18,9 @@ USE_EXTRA_THERMO = TRUE
 
 USE_CXX_EOS = TRUE
 
+USE_FORT_MICROPHYSICS = FALSE
+BL_NO_FORT=TRUE
+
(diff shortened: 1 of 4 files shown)


zingale:
test: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-06-19-002/test_aprox_rates_C.html

I don't understand why the original rate is different now :/

maxpkatz:
This test evaluates for multiple values of rho and T. Your change needs to be reset before the next call to the original rate.

maxpkatz:
That said, this would only work in serial execution, it wouldn't work in parallel. So maybe we need an optional argument to the rate routine to override.

zingale:
this passes now:
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-06-20/index.html

I've split the alternate c12ag rate off into it's own kernel so we don't have to worry about a race condition.

maxpkatz (approved):

PR #698, opened 2021-06-22
fix some uninitialized variables with Fortran nets from C++

zingale:


zingale:
with this change, there are no more valgrind issues

harpolea (approved):

PR #700, opened 2021-06-24
aprox13 15/∞

maxpkatz:


maxpkatz:
Note: this will significantly slow down the Jacobian evaluation (as it did in iso7). I need to fix this (in a separate PR) by computing as much of the Jacobian at compile time as possible, similar to the way we do it in the RHS.

zingale:
tests look reasonable: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-06-24-001/index.html

zingale (approved):

PR #706, opened 2021-07-10
convert the numerical Jacobian from header to source

zingale:
the inlining is slow to build on GPUs


zingale:
current `development` compilation time on groot (`make -j 20`) is:
```
real	35m7.487s
user	63m31.204s
sys	3m33.891s
```

zingale:
so far, this only shaves a little time off:
```
real	26m50.336s
user	54m52.933s
sys	3m22.424s
```

PR #688, opened 2021-06-10
some clean-ups of the simplified-SDC code paths

zingale:
in particular, this now zeros the rates if the temperature is out of bounds.

PR #708, opened 2021-07-12
accumulate changes for SDC testing

zingale:
This accumulates PRs #660, #688, #702 to test them in combination


zingale:
this now also has #710

zingale:
no longer needed

PR #710, opened 2021-07-13
greatly simplify how we do the Jacobian

zingale:


zingale:
this seems to work now.

Speed comparison: 10 steps of massive_star with SDC (8 cores of inf)  has runtime w/o initialize of 2010 s on development and 1891 with this PR.

This uses a lot less memory than the old implementation, so I expect the performance difference will be even more dramatic on GPUs.

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-07-14-002/index.html

that flame compilation failure is a test suite issue (that I've fixed)

zingale:
note: Maestro should be unaffected by this change -- I've made it's code path the same as before.  Until I have something that I can test with Maestro I am not going to update that.

integration/VODE/vode_jac_util_simplified_sdc.cpp
```diff
@@ -14,195 +14,122 @@ void jac_to_vode(const Real time, burn_t& state,
...
+        jac(SEDEN+1, SFS+n) = jac_react(net_ienuc, n) - jac_react(net_ienuc, net_ienuc) * eos_xderivs.dedX[n-1];
+        jac(SEINT+1, SFS+n) = jac(SEDEN+1, SFS+n);
     }
 
-    // d( d(rho e)/dt)/de
-    dRdw(SEINT+1, iwe) = state.rho * jac_react(net_ienuc, net_ienuc);
-
-    // d( d(rho E)/dt)/de
-    dRdw(SEDEN+1, iwe) = dRdw(SEINT+1, iwe);
+    jac(SEDEN+1, SEDEN+1) = 0.0_rt;
```
maxpkatz, line 83: If SEDEN = SEINT + K, how can this statement be true?

zingale:
I'm going to replace this with a simpler system

PR #699, opened 2021-06-24
Move identical particle factor out of rate

maxpkatz:
This is now computed automatically based on the reaction properties.

PR #721, opened 2021-07-25
aprox13 18/∞

maxpkatz:
Use auto species_rhs for d(rate)/dT in Jacobian

### DIFF
--- a/networks/aprox13/actual_rhs.H
+++ b/networks/aprox13/actual_rhs.H
@@ -1278,178 +1278,6 @@ void ener_gener_rate(T const& dydt, Real& enuc)
 
 }
 
-// Evaluates the right hand side of the aprox13 ODEs
-AMREX_GPU_HOST_DEVICE AMREX_INLINE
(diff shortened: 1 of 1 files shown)


maxpkatz:
This will cause a test suite diff. The reason is that there was a bug in #699: I didn't remember that this part of aprox13 hadn't yet been converted to the generic species_rhs functionality, so I didn't remove the identical particle factor from this rhs() function in aprox13 to compensate. As a result that did actually substantially change the d/dE terms in the Jacobian for aprox13 as indicated in test_rhs, I just didn't notice it at the time: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-07-22-002/test_rhs_aprox13.html. This change gets the identical particle factor right automatically.

maxpkatz:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-07-25/index.html

As indicated, the J_*_E terms that include reactions with identical particles (e.g. triple alpha, O16+O16->S32) have large diffs, the other ones are roundoff.

zingale (approved):

PR #726, opened 2021-07-29
Add a constexpr version of screen5

maxpkatz:
An overload of screen5 is added that accepts as argument a screen_factors_t, as well as a constexpr function that can do a compile-time evaluation of these factors (depends on the new support added in #725). This allows the constant screening factors to be evaluated inline in the screening routine at compile time without needing the separate set_up_screening_factors() step.


zingale:
all tests pass: 
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-07-29-002/index.html

zingale (approved):

PR #727, opened 2021-07-29
aprox13 19/∞

maxpkatz:
Switch to constexpr screen factors evaluation


maxpkatz:
I expect roundoff differences in the RHS since gcem's compile-time evaluation of log and pow can be slightly different from what we would get at runtime.

zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-07-29-004/index.html

RHS diffs are roundoff

zingale (approved):

PR #730, opened 2021-07-30
Work around lack of long double support on Power9

maxpkatz:


zingale (approved):

zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-07-30-002/index.html

maxpkatz:
Those diffs look OK to me. The effect of this change is that we are now changing the format of the constant value of `e` in the code from long double to double, which could result in different truncation/roundoff in the last digit, and this value of `e` is used in constructing some of the compile-time evaluations, so the differences at that level in the RHS don't surprise me.

PR #731, opened 2021-07-31
Compute some Jacobian terms at compilation time

maxpkatz:
The current species Jacobian implementation inefficiently loops through every rate at runtime, computes whether it is used in a given Jacobian element, and is then computed. But we know at compile time which rates are used, so we can limit the calculation to only those rates, and skip a lot of runtime branching. This has already been done for the RHS and now the Jacobian is treated using a similar approach.


maxpkatz:
I sometimes use a setup like this to test performance:

```
./main3d.gnu.CUDA.ex inputs_aprox13 amrex.the_arena_init_size=0 n_cell=64 max_grid_size=1024 unit_test.dens_min=1.e7 un
it_test.dens_max=1.e7 unit_test.temp_min=1.e8 unit_test.temp_max=1.e8
```

This means that every zone does basically the same thing and there's only a few RHS calls per zone, so this is a useful way of checking whether there are any inefficiencies in the code implementation (separately from the known algorithmic problem of differing number of ODE steps in the general case). For this setup on V100 this change improves performance by over 2x (183 ms -> 83 ms), and instead of the Jacobian construction being the bottleneck, the linear algebra becomes the bottleneck as expected.

zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-07-30-002/index.html

zingale:
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-07-31-002/index.html

zingale:
is this still something we want?

maxpkatz:
Closing, this referred to an older version of the code

PR #732, opened 2021-07-31
Fix aprox13 actual_matrix references to net_itemp

maxpkatz:


zingale:
I don't think we have a test for this -- how do I compile with this to test it?

maxpkatz:
USE_NETWORK_SOLVER=TRUE

But it does not compile right now which is what I am fixing

zingale (approved):

PR #738, opened 2021-08-04
redo the numerical Jacobian (again)

zingale:
now we difference in terms of T, X_k, but convert to e, X_k

### DIFF
--- a/integration/utils/numerical_jacobian.H
+++ b/integration/utils/numerical_jacobian.H
@@ -16,7 +16,11 @@
 /// in terms of X.
 ///
 /// Note: even for the various flavors of SDC, this is the form we
-/// want (we call this dR/dw), and we will convert to 
+/// want (we call this dR/dw)
+///
+/// Even though we have e as an independent variable, we will
+/// difference in terms of X and T and then convert the Jacobian
+/// elements to be in terms of X and e
 
 struct jac_info_t {
     Real h;
@@ -40,6 +44,9 @@ void numerical_jac(burn_t& state, const jac_info_t& jac_info, JacNetArray2D& jac
     YdotNetArray1D ydotp;
     YdotNetArray1D ydotm;
 
+    // note: we assume that coming into the numerical jacobian that
+    // state is thermodynamically consistent (T matches e)
+
     burn_t state_delp = state;
 
     // default -- plus convert the dY/dt into dX/dt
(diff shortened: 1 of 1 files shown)


integration/utils/numerical_jacobian.H
```diff
@@ -105,37 +108,75 @@ void numerical_jac(burn_t& state, const jac_info_t& jac_info, JacNetArray2D& jac
...
+    eos_state.e = state.e;
+    for (int n = 0; n < NumSpec; n++) {
+        eos_state.xn[n] = state.xn[n];
+    }
+#ifdef NSE_THERMO
+    // make the aux data consistent with the state X's
+    set_nse_aux_from_X(eos_state);
+#endif
+
+    eos(eos_input_re, eos_state);
```
maxpkatz, line 161: If it is true that the state is thermodynamically consistent coming into this call, then we shouldn't need this EOS evaluation, state.cv should already be correct if state.T is correct.

maxpkatz: Actually, I suppose we still need it for composition_derivatives since dedA and dedZ are not in burn_t by default.

zingale: indeed, I didn't fully check that bit, but I suspect you are right.  I just need to be able to compute the composition derivatives.

integration/utils/numerical_jacobian.H
```diff
@@ -105,37 +108,75 @@ void numerical_jac(burn_t& state, const jac_info_t& jac_info, JacNetArray2D& jac
...
+    eos_xderivs_t eos_xderivs = composition_derivatives(eos_state);
+
+    // now correct the energy column
+
+    for (int m = 1; m <= neqs; m++) {
+        jac(m, net_ienuc) *= 1.0_rt / eos_state.cv;
+    }
+
+    // now correct the species derivatives
+    // this constructs dy/dX_k |_e = dy/dX_k |_T - e_{X_k} |_T dy/dT / c_v
```
maxpkatz, line 172: Remind me why this is a minus sign and not a plus sign?

zingale: imagine ω_k = ω_k(ρ, T(ρ, X_k, e), X_k)

∂ω_k/∂X_j |_e = ∂ω_k/∂X_j |_T + ∂ω_k/∂T ∂T/∂X_k |_{ρ, e}

then 

de = ∂e/∂T |_{ρ, X_j} dT + ∂e/∂ρ |_{T, X_j} dρ + ∂e/∂X_j dX_j = 0 (since de = 0 in the derivative above)

taking constant density then gives:

∂T/∂X_j |_{ρ, e} = - ∂e/∂X_j  / c_v

zingale:
Microphysics tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-08-05-002/index.html

zingale:
Castro test: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-08-05-002/massive_star-sdc.html

maxpkatz (approved):

PR #739, opened 2021-08-05
only check for NSE bailout after 10 steps

zingale:
VODE sometimes does some initial exploration that is rejected, but may
trigger the NSE bailout, resulting in a very bad state with huge NSE
energy release.  Now we instead don't check for the NSE bailout until after
10 steps have been taken.  The number 10 is arbitrary, but seems sufficient
for VODE to have settled down.

### DIFF
--- a/integration/VODE/actual_integrator_simplified_sdc.H
+++ b/integration/VODE/actual_integrator_simplified_sdc.H
@@ -215,7 +215,7 @@ void actual_integrator (burn_t& state, Real dt)
 #endif
         } else {
 #ifndef AMREX_USE_GPU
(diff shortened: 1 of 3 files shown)


zingale:
this has an added bonus that it might in general be faster, since we avoid the first N checks on states that don't go into NSE (and therefore the EOS calls)

zingale:
Microphysics tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-08-05-001/index.html

zingale:
Castro tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-08-05-001/index.html

maxpkatz (approved):

PR #742, opened 2021-08-12
Copy cv from Burn State when going from c++ to fortran

biboyd:
Currently when going from C++ to Fortran, values from the burn state like T, rho, etc. are copied over but cv does not. This can cause an error when `integration/utils/temperature_integration.f90:temperature_rhs()` is called but cv has not yet been initialized.

This fixes that problem by copying over the cv when transitioning from c++ code to fortran.

Additionally some path names to some URCA reaction tables needed to be updated.

### DIFF
--- a/networks/ignition_reaclib/URCA-simple/actual_rhs.F90
+++ b/networks/ignition_reaclib/URCA-simple/actual_rhs.F90
@@ -134,7 +134,6 @@ contains
     use extern_probin_module, only: do_constant_volume_burn, disable_thermal_neutrinos
     use burn_type_module, only: net_itemp, net_ienuc, neqs
     use sneut_module, only: sneut5
-    use temperature_integration_module, only: temperature_rhs
 
     implicit none
 
@@ -171,9 +170,6 @@ contains
 
     ! Append the energy equation (this is erg/g/s)
     ydot(net_ienuc) = enuc - sneut
-
-    ! Append the temperature equation
-    call temperature_rhs(state, ydot)
 
   end subroutine actual_rhs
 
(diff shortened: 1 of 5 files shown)


biboyd:
tag @dwillcox

zingale (approved):

zingale:
This looks good to me.  But I think that we need to think about whether Maestro should be integrating h instead of e, and therefore need c_p instead of c_v.  We should write this out on paper to see what it implies.

networks/ignition_reaclib/URCA-simple/table_rates.F90
```diff
@@ -55,7 +55,7 @@ subroutine init_tabular()
     num_rhoy_j_na23_ne23 = 152
     num_vars_j_na23_ne23 = 6
     num_header_j_na23_ne23 = 6
-    rate_table_file_j_na23_ne23 = trim("23Na-23Ne_electroncapture.dat")
+    rate_table_file_j_na23_ne23 = trim("23Na-23Ne_electroncapture.network.dat")
```
zingale, line 58: I think this change needs to be made in pynucastro, since this file is generated from that, and if we regenerate this, this change will be lost.

biboyd:
I believe this should correctly fix the bug since we don't use the output of temperature_rhs() anyways.

networks/ignition_reaclib/URCA-simple/actual_rhs.F90
```diff
@@ -241,7 +237,6 @@ subroutine actual_jac(state, jac)
     use burn_type_module, only: net_itemp, net_ienuc, neqs, njrows, njcols
     use extern_probin_module, only: disable_thermal_neutrinos
     use sneut_module, only: sneut5
-    use temperature_integration_module, only: temperature_jac
     use jacobian_sparsity_module, only: get_jac_entry, set_jac_entry, set_jac_zero
```
zingale, line 240: these files are automatically generated, so I think that we should remove the lines from the relevant part of pynucastro

zingale:
I think that there's actually an issue with the Jacobian too -- we want it to be dydot/de which we can compute from dydot/dT, but I don't think that's what we are doing in the wrapper.  This again is something to handle in pynucastro, I think.

zingale:
this is fixed by the C++ port in #809

PR #745, opened 2021-08-13
create a temp burn_state for NSE bailout

zingale:
something in the update of the burn_state to check for NSE causes
the integration to fail at times when running with SDC+NSE.  This
works around that.  This can be tested with the new burn_cell_sdc
problem and the inputs_aprox19_failure inputs.


zingale:
blocked by #744

PR #751, opened 2021-08-27
add a slightly more extensive nova network

zingale:
this includes all of the nuclei that Ami and Jordi use


aisclark91 (approved):
This reaction network solves the problem and incorporates new options for CNO explosions.

PR #755, opened 2021-08-31
if we bailout due to NSE, properly set the time

zingale:


zingale:
the logic there is that when we exit VODE after this bailout, in the `actual_integrator_simplified_sdc.H` we use `vstate.t` to make the state consistent.  but it was not being set properly (at all?), since that is only really set at the end.

so I think that when we finish the burn via NSE, we were doing the entire timestep's advection + whatever advection we had already done up to that point in VODE-SDC

for Strang, I think it might just be getting the energy a bit wrong, but it doesn't mess up the advection like it does with SDC.

maxpkatz (approved):

PR #757, opened 2021-09-02
remove Fortran support from test_eos

zingale:
also get rid of the probin's


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-09-02-002/index.html

I think that the failures are for quantities that are not filled for these EOSes?

maxpkatz (approved):

PR #767, opened 2021-09-12
iso7 28/∞

maxpkatz:
Adds a function to automatically compute density dependence of tabulated rates.

The Ca40 + He4 <-> Ti44 rate is not dealt with yet since it doesn't have an entry in rhs_data().

### DIFF
--- a/interfaces/rhs_utilities.H
+++ b/interfaces/rhs_utilities.H
@@ -210,14 +210,14 @@ constexpr int density_exponent_reverse ()
 
     int exponent = 0;
 
-    if constexpr (rhs_data.species_C >= 0) {
-        exponent += rhs_data.exponent_C;
-    }
     if constexpr (rhs_data.species_D >= 0) {
         exponent += rhs_data.exponent_D;
     }
     if constexpr (rhs_data.species_E >= 0) {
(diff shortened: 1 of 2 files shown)


zingale:
tests:  http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-09-12/index.html

interfaces/rhs_utilities.H
```diff
@@ -173,6 +173,60 @@ constexpr int num_rhs ()
     return RHS_impl::num_rhs_impl<species>(std::make_integer_sequence<int, Rates::NumRates>{});
 }
 
+// Calculate the density dependence term for tabulated rates. The RHS has a term
+// that goes as rho**(exp_A + exp_B + exp_C) / rho (the denominator is because the
+// LHS is X, not rho * X).
```
zingale, line 178: "the denominator is..." what? I'm on the edge of my seat...

zingale (approved):

interfaces/rhs_utilities.H
maxpkatz: I meant it in the sense of "the denominator *exists* because..."

PR #769, opened 2021-09-15
Add a templating equivalent to EXTRA_THERMO

maxpkatz:
The functionality is demonstrated in Helmholtz.

In order to make test_eos work with USE_EXTRA_THERMO=FALSE, the code that actually does the work needs to be in a templated function (since if constexpr only skips checking the validity of the false condition inside template functions).


maxpkatz:
Note: this does not yet actually remove EXTRA_THERMO, it just makes some parts of Microphysics agnostic as to whether the eos_t object has dpdA, etc., in the type. When this is complete (in both Microphysics and Castro/Maestro) then we will be easily able to define a new type inheriting from eos_t that has the extra thermo components and then everything should just work without needing the ifdef.

zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-09-15-002/index.html

PR #771, opened 2021-09-15
iso7 29/∞

maxpkatz:
Do density dependence of analytic rates using new function. As with #767 we aren't yet doing this with the Ca40<->Ti44 rate since that requires additional work.


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-09-17-001/index.html

zingale (approved):

PR #772, opened 2021-09-17
add a fixed T to SDC to support drive_initial_convection

zingale:


interfaces/burn_type.H
```diff
@@ -134,6 +134,11 @@ struct burn_t
   int sdc_iter;
   int num_sdc_iters;
 
+  // for drive_initial_convection, we will fix T during the
+  // integration to a passed in value.  We will interpret a positive
+  // T_fixed as setting this feature.
+  Real T_fixed;
```
maxpkatz, line 140: With this we have to make sure T_fixed is set to 0 even when it's not used.

zingale: right.  But it only affects SDC.  

Does it work to initialize it in the struct?

maxpkatz (approved):

PR #775, opened 2021-09-30
General null gpu

ajnonaka:
Give general_null the option to compile/run on GPU.


zingale (approved):

zingale:
I guess the reason Castro doesn't have this issue is because we don't compile with `USE_REACT=TRUE` when running Castro.

PR #779, opened 2021-10-05
Add concept of extra species that do not directly react

maxpkatz:
We can now define additional auxiliary species to the network that aren't directly reacting but are used in intermediate rate calculations (such as the Ca40 + He4 -> Ti44 intermediate rate used in iso7 which is used to construct the overall Si28 -> Ni56 rate).

### DIFF
--- a/networks/general_null/network_header.template
+++ b/networks/general_null/network_header.template
@@ -17,6 +17,10 @@
 
 constexpr int NumSpec = @@NSPEC@@;
 
+#define NUM_EXTRA_SPECIES @@NEXTRASPEC@@
+
+constexpr int NumSpecExtra = NUM_EXTRA_SPECIES;
+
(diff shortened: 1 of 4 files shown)


maxpkatz:
Looking forward, my goal is to be able to automate as much as possible of the rate construction in actual_rhs.H, which means minimizing the number of hardcoded quantities. In #771 I was unable to automatically compute the density dependence of the Ca40 -> Ti44 rate because this reaction does not exist in the rhs_data construct that determines which reactions are in the network. Adding Ca40_He4_to_Ti44 to that is tricky because we don't want that rate to actually go into the construction of the RHS; this rate only exists temporarily during rate evaluation, and it is Si28_7He4_to_Ni56 that is the real rate. I am not yet sure how best to actually add this intermediate rate but any sort of automation would ultimately require some notion of the species as entities that have an A and Z, so this moves us forward.

PR #783, opened 2021-10-08
iso7 30/∞

maxpkatz:
We now have automatic tabulated rate density factor for Ca40->Ti44. The rate is added to rhs_data, but to exclude it from the actual RHS construction we make sure that we ignore reactions involving species whose ID is larger than NumSpec, which must be "extra" species.


maxpkatz:

PR #785, opened 2021-10-14
Remove dratdumdy arrays from aprox19

maximumcats:
The functionality was removed in #686, this just removes the unnecessary dead weight memory.


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-10-14-001/index.html

PR #790, opened 2021-10-16
Add a constexpr for loop for linear algebra

maxpkatz:


maxpkatz:

PR #791, opened 2021-10-16
Implement RHS using a compile-time for loop

maxpkatz:
The current method for constructing the RHS involves creating an array of all terms which participate in the reaction and then summing them together. This is suboptimal for two reasons: the construction of the intermediate array is no longer necessary (since we're not using esum anymore), and the code is quite hard to understand. But with C++17 we can effectively get compile-time/constexpr for loops using the method described [here](https://artificial-mind.net/blog/2020/10/31/constexpr-for). The main change is then that the body of the for loop is a lambda whose argument is the for loop index. As a result we can write an intuitive RHS construction which loops over species and then rates and adds a given term to the RHS if that species is involved in that rate.

The primary drawback is that gcc < 9 does not compile this code, so gcc >= 9 would become the de facto minimum compiler version for Castro and Maestro. If this became a problem, we could work around it by writing a version of `rhs_term()` that does not depend on any compile-time evaluation. It would be much slower at runtime relative to the current code, though -- some benchmarking suggests it would make test_react for iso7 take more than 2x the current time.

### DIFF
--- a/interfaces/rhs_utilities.H
+++ b/interfaces/rhs_utilities.H
@@ -58,81 +58,15 @@ constexpr int factorial (int n)
     }
 }
 
-// Base case for recursion.
-template<int species>
-AMREX_GPU_HOST_DEVICE AMREX_INLINE
-constexpr int add_nonzero_rates ()
+// Implementation of "constexpr for" based on
+// https://artificial-mind.net/blog/2020/10/31/constexpr-for
+template<auto I, auto N, class F>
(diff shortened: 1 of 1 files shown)


maxpkatz:
Note for review: it is easier to directly focus on understanding what the new implementation of `species_rhs()` looks like, instead of focusing on the diff itself.

maxpkatz:

PR #802, opened 2021-10-23
aprox network rewrite

maxpkatz:


zingale:
test suite: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2022-02-19/index.html

Note: simplified-SDC doesn't work with this -- I think we need to ifdef out the linear algebra changes for now if we are not STRANG

zingale:
here are the Castro tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2022-02-19/index.html

the errors look reasonable, and most problems using these nets show nice speed ups (the one exception is wdmerger_collision, but that is built with `DEBUG=TRUE`)

zingale:
it looks like `interfaces/rhs_utilities.H` can be removed with this PR

networks/rhs.H
```diff
@@ -0,0 +1,1604 @@
...
+
+        if (data.additional_reaction_3 >= 1) {
+            rate_is_intermediate[data.additional_reaction_3 - 1] += 1;
+        }
+    }
+
+    // Now count up all intermediate rates.
+    int count = 0;
+
+    for (int rate = 1; rate <= Rates::NumRates; ++rate) {
```
zingale, line 183: doesn't this double count an intermediate rate that is an additional rate to more than one rate?

networks/rhs.H
```diff
@@ -0,0 +1,1604 @@
...
+
+        Real sc, scdt;
+        actual_screen5(pstate, scn_fac, sc, scdt);
+
+        if constexpr (data.screen_forward_reaction == 1) {
+            frdt = frdt * sc + fr * scdt;
+            fr   = fr * sc;
+        }
+
+        if constexpr (data.screen_reverse_reaction == 1) {
```
zingale, line 552: is this right?  For A + B -> C+ D, the reverse rate screening should use aion and zion from C and D

networks/rhs.H
```diff
@@ -0,0 +1,1604 @@
...
+void apply_branching (Real& fr, Real& frdt, Real& rr, Real& rrdt)
+{
+    constexpr rhs_t data = rhs_data(rate);
+
+    if constexpr (data.forward_branching_ratio != 1.0_rt) {
+        fr *= data.forward_branching_ratio;
+        frdt *= data.forward_branching_ratio;
+    }
+
+    if constexpr (data.forward_branching_ratio != 1.0_rt) {
```
zingale, line 636: should this be data.reverse_branching_ratio ?

networks/rhs.H
```diff
@@ -0,0 +1,1604 @@
...
+{
+    constexpr rhs_t data = rhs_data(rate);
+
+    // First, compute the Y * rate component of both the forward and
+    // reverse reactions, which is the same regardless of which species
+    // we're producing or consuming.
+
+    Real forward_term = fr;
+
+    if constexpr (data.species_A >= 0) {
```
zingale, line 765: can it really be == 0?  shouldn't this just be > 0?

networks/rhs.H
```diff
@@ -0,0 +1,1604 @@
...
+        }
+
+        if constexpr (data.species_B >= 0) {
+            Real Y_B = state.xn[data.species_B-1] * aion_inv[data.species_B-1];
+
+            constexpr int exponent = data.exponent_B;
+
+            constexpr int exp = (spec2 == data.species_B) ? exponent - 1 : exponent;
+
+            static_assert(exp <= 3);
```
zingale, line 1006: based on the assert in rhs_term, shouldn't this assert ensure that exp <= 2?

zingale: oh, only if we are differenting wrt it.  I see the logic.  nevermind

networks/rhs.H
maxpkatz: Yes, fixed

maxpkatz:
> it looks like `interfaces/rhs_utilities.H` can be removed with this PR

Done

networks/rhs.H
maxpkatz: I'm anticipating a future change where we do zero-indexing on the species, less code we'll have to change at that time.

networks/rhs.H
maxpkatz: This was definitely an issue I was concerned about, so my approach is to start with no reactions being intermediate (the counter is zero for all rates), and then we loop through all rates and increment the counter each a reaction is intermediate to some other reaction. It will be the case in general that some have a counter value > 1, but in the second loop where we actually count the number of intermediates, we're only checking for counter value > 0.

maxpkatz:
> Note: simplified-SDC doesn't work with this -- I think we need to ifdef out the linear algebra changes for now if we are not STRANG

I think this was just because I had forgotten to update numerical_jacobian.H to use the new interfaces when appropriate. This is now fixed.

zingale:
I'll retest

zingale (approved):

PR #803, opened 2021-10-29
use a more general strategy for linking net tables

zingale:


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-10-29-001/index.html

maxpkatz (approved):

PR #804, opened 2021-10-29
bail if the primary species isn't defined

zingale:
we were doing illegal memory accesses otherwise


zingale:
tests are good: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-11-19-001/index.html

maxpkatz (approved):

PR #805, opened 2021-10-30
fix the Fortran Jacobian wrapper for J_E_E

zingale:
we were not accessing the energy component


zingale:
only one test currently uses the wrapper:
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-10-30/test_react_C-nova-VODE.html

maxpkatz (approved):

PR #806, opened 2021-10-30
convert the Urca pynucastro net over to C++

zingale:


zingale:
@biboyd this should fix a few issues with the Urca net.  Please test this out and report back here in the issue if you think this is accurate.

zingale:
this is contained in #809

PR #807, opened 2021-10-31
convert nova, nova2, subch, and subch2 to C++

zingale:
This uses the latest changes in pynucastro


zingale:
this is contained in #809

PR #809, opened 2021-11-03
convert all pynucastro nets to C++

zingale:


zingale:
first attempt at testing:
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-11-04-007/index.html

the nova build failure is because I need to turn off Fortran.
The J_E_* diffs are because of physical constants.  I'll fix that shortly and rerun.

zingale:
curiously there are 3 different values of erg <-> eV in Microphysics.  The Fortran nets use their own instead of getting it from our constants already.  I am not sure I want to add yet a third value to C++ just to make these errors go down.

zingale:
updated tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-11-22-003/index.html

maxpkatz (approved):

PR #810, opened 2021-11-04
add inputs for nova and ECSN

zingale:
also remove tmax -- it is not used for this test


zingale:
this will let me cover most of the pynucastro nets in the unit testing now.

PR #814, opened 2021-11-16
add inputs that causes SDC integration fail with step rejection

zingale:
if we uncomment out the logic in vode_dvstep.H that rejects a step if
the change in a species from one step to the next is too large, then
this inputs fails.


zingale:
to reproduce, build `burn_cell_sdc` on the `pynucastro-cxx2` branch as:

```
make NETWORK_DIR=subch2
```

then uncomment out the lines in `vode_dvode.H` that reject based on the change in species from one step to the next.

Note: if you change the parameters to allow jumps of 4 instead of 2, then this works.

maxpkatz (approved):

PR #815, opened 2021-11-17
store y_save before attempting any steps

zingale:
we were storing it in the step trial loop, which meant if there
were successive failures, then we would use the old failed state
as the reference to check the increase in species against.  This
make the species step rejection stuff not work reliably.

This also changes SDC to limit based on X instead of rho X

### DIFF
--- a/integration/VODE/vode_dvstep.H
+++ b/integration/VODE/vode_dvstep.H
@@ -134,6 +134,26 @@ int dvstep (burn_t& state, dvode_t& vstate)
     // dvset is called to calculate all integration coefficients.
     // RC is the ratio of new to old values of the coefficient H / EL(2) = h / l1.
 
+    // Save the value of y before calling the solver.
+    // We will use this saved value to determine if any
+    // constraints were violated in the update.
+
+    Array1D<Real, 1, VODE_NEQS> y_save;
+#ifdef STRANG
+    for (int i = 1; i <= VODE_NEQS; ++i) {
+        y_save(i) = vstate.y(i);
+    }
+#endif
+#ifdef SIMPLIFIED_SDC
+    Real rho_old = state.rho_orig + TOLD * state.ydot_a[SRHO];
+    for (int i = 1; i <= VODE_NEQS; ++i) {
+        y_save(i) = vstate.y(i);
+    }
+    for (int i = 1; i <= NumSpec; ++i) {
+        y_save(SFS+i) /= rho_old;
+    }
+#endif
+
     while (true) {
 
         vstate.tn += vstate.H;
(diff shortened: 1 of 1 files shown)


zingale:
when I run the case in #814 

this is the output I get from the species step rejection:

```
rejecting step based on species 7 from 1.5010126834e-05 to 5.98853565043e-05
rejecting step based on species 7 from 1.5010126834e-05 to 0.0147124994758
rejecting step based on species 7 from 1.5010126834e-05 to 0.00318753985433
rejecting step based on species 7 from 1.5010126834e-05 to 0.000338029355527
rejecting step based on species 7 from 1.5010126834e-05 to 4.73708672645e-05
rejecting step based on species 7 from 2.14833226375e-05 to 5.38414440912e-05
rejecting step based on species 7 from 2.14833226375e-05 to 0.0168619846071
rejecting step based on species 7 from 2.14833226375e-05 to 0.00309206662756
rejecting step based on species 7 from 2.14833226375e-05 to 0.000333940345273
rejecting step based on species 7 from 2.14833226375e-05 to 5.27840578425e-05
```

so this shows that there are 2 steps that fail multiple times, which is why the old version of this `y_save` had so much trouble.

zingale:
reaction tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-11-17-001/index.html

integration/VODE/vode_dvstep.H
```diff
@@ -253,12 +267,16 @@ int dvstep (burn_t& state, dvode_t& vstate)
 
 
             // this doesn't seem to work well with SDC
```
maxpkatz, line 269: This comment should be removed

integration/VODE/vode_dvstep.H
```diff
@@ -228,10 +240,12 @@ int dvstep (burn_t& state, dvode_t& vstate)
             // factor in a given step (as long as their initial abundance was large enough
             // for us to worry about).
 
-            if (std::abs(y_save(i)) >= atol_spec &&
+            if (std::abs(y_save(i)) > vode_increase_change_factor * atol_spec &&
                 (std::abs(vstate.y(i)) > vode_increase_change_factor * std::abs(y_save(i)) ||
                  std::abs(vstate.y(i)) < vode_decrease_change_factor * std::abs(y_save(i)))) {
+                std::cout << "rejecting step based on species " << i << " from " << y_save(i) << " to " << vstate.y(i) << std::endl;
```
maxpkatz, line 246: This print should be guarded out on GPUs. Also it will generate a lot of output so I'm not sure if it's wise to have it on by default in CPU runs either. Perhaps we should have some sort of MICROPHYSICS_DEBUG macro for info like this.

zingale: I like a MICROPHYSICS_DEBUG.  I'll add that

integration/VODE/vode_dvstep.H
zingale: done

integration/VODE/vode_dvstep.H
zingale: double guarded with AMREX_USE_GPU and MICROPHYSICS_DEBUG

zingale:
I just updated this to not enable the SDC checks yet, but we will do that in #818

maxpkatz (approved):

PR #821, opened 2021-12-01
synchronize the strang and SDC step rejection logic

zingale:
also boost the threshold


zingale:
this seems to work really well for both the Strang and SDC subchandra runs.

maxpkatz (approved):

PR #822, opened 2021-12-02
remove Fortran code paths from unit tests

zingale:
these were not used and we will be removing the Fortran implementations
soon anyway.


zingale:
all tests pass except for cond -- that was testing the Fortran version and now it does the C++ version and diffs are roundoff as to be expected:
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-12-02-003/index.html

unit_test/test_screening/main.cpp
```diff
@@ -117,22 +106,17 @@ void main_main ()
...
 
     amrex::Vector<std::string> names;
     get_varnames(vars, names);
 
-    // Fortran test
-
-    init_variables_F();
-
     // time = starting time in the simulation
     Real time = 0.0;
```
aisclark91, line 121: Suggestion 1: Add the specifier _rt to all the constants and cast n_cell to Real below

aisclark91 (commented):
I think it could be a good idea to add the gcem library instead of the cmath library to perform the log10 computations.

zingale:
I think gcem only makes sense when we are evaluating things at compile time, not run time.

zingale:
retested with casting and `_rt`:
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-12-08-001/index.html

the screening diffs went away.

aisclark91:
If the differences went away I will approve it.

aisclark91 (approved):

PR #825, opened 2021-12-04
do a better job at cleaning up linked files

zingale:
this now cleans up the pynucastro files and helm in the _extern stub
this way application codes pick up these rules automagically


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-12-03-001/index.html

maxpkatz (approved):

PR #826, opened 2021-12-04
remove .F90 files for nets with C++ implementation

zingale:
also breakout was a null net, so merge that into general_null


zingale:
this works, but depends on #822

zingale:
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-12-04-001/index.html

Issue #829, opened 2021-12-14
general_null doesn't handle naux correctly


zingale:
The `general_null` network has `NAUX_NETWORK` with `naux := 0` hardcoded.  This means that it doesn't work with null networks that define auxiliary quantities (like `breakout.net`).

PR #836, opened 2021-12-17
eliminate NAUX_NETWORK in favor of auto-detection

zingale:
instead of hardcoding the number of aux quantities in the network, we
now determine it at compile time by parsing the .net file via the
new get_naux.py script.  This also separates some of the commonly
used network property file stuff into a separate routine.


zingale:
this closes #829

zingale:
Castro tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-12-17-001/index.html

maxpkatz (approved):

Issue #833, opened 2021-12-15
nonaka_plot is not capturing jacobian=3 RHS evals


zingale:
For the nonaka plot, we are catching all of the RHS evaluations for both the integration and computing the jacobian = 2 numerical Jacobian, since that all goes through the `rhs()` wrapper.  But for jacobian = 3, we are not.

Maybe we want to exclude the jacobian evaluations from the nonaka plot? and only show the RHS evals in the course of doing the integration?

PR #838, opened 2021-12-18
don't output state to nonaka plot file for Jacobian evaluation

zingale:
we should just store the RHS evals used in the advance, and not for
computing the numerical Jacobian


zingale:
this addresses #833 by removing all Jacobian evals from the output

harpolea (approved):

PR #839, opened 2021-12-28
Fix Fe54 Jacobian terms in aprox19

maxpkatz:


maxpkatz:
This indexing bug appears to have been present since the original addition of aprox19 in commit 640f4aea3768993e1eb55ece34fd749a937b7854. However it is not an issue in the current upstream aprox19 code on Frank Timmes' website.

zingale (approved):

PR #840, opened 2021-12-28
Remove implicit d/dP terms in aprox19 and aprox21 Jacobians

maxpkatz:
In both aprox19 and aprox21 there is the reaction sequence Fe52(a,p)Co55(p,g)Ni56. Co55 is an equilibrium isotope in this reaction chain, and its equilibrium abundance depends on both this reaction sequence as well as the reaction Fe54(p,g)Co55. Since there are two contributions to the effective y(Co55), the rate ends up being somewhat complicated, but the RHS term ends up simply looking like `y(Fe52) * y(He4) * y(P) * rate`. The Jacobian terms consequently have a d/dP entry. However, although the proton abundance plays a role in determining the rate, the reaction sequence doesn't actually consume or produce a net proton, so it is more appropriate to think of y(P) as being part of the rate rather than part of the RHS. If the y(P) is absorbed into the rate, then this term should not appear in the Jacobian due to the same reason #686 was done: in the upcoming network rewrite, it will not be straightforward (or desirable) to represent any d(rate)/dy term. So this PR proposes removing the term on the basis that it will eventually be removed anyway, and this will make the network rewrite PR easier to evaluate when comparing test_rhs and test_react.

### DIFF
--- a/networks/aprox19/actual_rhs.H
+++ b/networks/aprox19/actual_rhs.H
@@ -1896,9 +1896,9 @@ void dfdy_isotopes_aprox19 (Array1D<Real, 1, NumSpec> const& y,
     b(1)  =  2.0e0_rt * y(Fe54) * y(P) * ratdum(ir5f54);
     b(2)  =  0.0;
     b(3)  =  0.0;
-    b(4)  = -y(Fe52) * y(He4) * ratdum(ir7f54);
+    b(4)  =  0.0;
     b(5)  =  0.0;
-    b(6)  =  y(Ni56) * ratdum(ir8f54);
+    b(6)  =  0.0;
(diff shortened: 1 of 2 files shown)


maxpkatz:
Test suite: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2021-12-28-001/index.html

As intended there are diffs in the d/d(proton) terms for a few species. For the aprox19 test_react, the run finishes in about the same time with about the same number of average RHS evaluations. Looks like we don't have an aprox21 test_react run that uses the analytic Jacobian though.

zingale:
this seems consistent with the other approximations we've done in the Jacobians.

zingale (approved):

PR #861, opened 2022-01-03
remove the last fortran bits

zingale:
note: test_parameters was intended to test passing parameters from
C++ to/from Fortran, so it is no longer needed


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2022-01-03-001/index.html

maxpkatz (approved):

PR #866, opened 2022-01-05
Update documentation on reaction ODE system

maxpkatz:
These changes remove references to Fortran, and update the reaction system to the one that is based on e instead of T. (Along with this, we also now point out that this e is the real internal energy instead of the tracer variable it was before.)

### DIFF
--- a/sphinx_docs/source/integrators.rst
+++ b/sphinx_docs/source/integrators.rst
@@ -8,37 +8,29 @@ Reaction ODE System
 The equations we integrate to do a nuclear burn are:
 
 .. math::
-   \frac{dX_k}{dt} = \omegadot_k(\rho,X_k,T)
+   \frac{dX_k}{dt} = \dot{\omega}_k(\rho,X_k,T)
    :label: eq:spec_integrate
 
 .. math::
-   \frac{d\enuc}{dt} = f(\dot{X}_k)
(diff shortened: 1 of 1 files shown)


sphinx_docs/source/integrators.rst
```diff
@@ -8,46 +8,38 @@ Reaction ODE System
...
-   \frac{dT}{dt} =\frac{\edot}{c_x} .
-   :label: eq:temp_integrate
-
-Here, :math:`X_k` is the mass fraction of species :math:`k`, :math:`\enuc` is the specifc
-nuclear energy created through reactions, :math:`T` is the
-temperature [1]_ , and :math:`c_x` is the specific heat for the
-fluid. The function :math:`f` provides the energy release based on the
+Here, :math:`X_k` is the mass fraction of species :math:`k`, :math:`e` is the specifc
+nuclear energy created through reactions. Also needed are density :math:`\rho`,
+temperature :math:`T`, and the specific heat. The function :math:`f` provides the energy release based on the
```
zingale, line 20: do we actually need the specific heat?

maxpkatz: Just for the analytic Jacobian.

zingale (approved):

PR #880, opened 2022-02-07
fix some array sizes in sneut

zingale:


zingale:
tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2022-02-07-001/index.html

maxpkatz (approved):

PR #888, opened 2022-02-23
add some more nuclei to better represent C12+C12, C12+O16, and O16+O16

zingale:
this uses the new pynucastro validate() method to identify rates that we might be missing.  This brings the number of nuclei from 28 to 33.


zingale:
subch2

^ that's a representation of the network

maxpkatz (approved):

zingale:
this net is more complicated than needed and both Strang and SDC have troubles integrating it.  I'm going to do something a little simpler.

PR #900, opened 2022-03-08
add a generic integrator_size.H

zingale:
this defines the size of the ODE system we integrate and knows
about the different coupling methods.


zingale:
tests are good: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2022-03-08-002/index.html
the sdc failure is because I hadn't merged the latest development in.

maxpkatz:
Just for future-proofing and consistency with other bits of Microphysics, can we call this `integrator_data.H`?

zingale:
renamed

integration/integrator_data.H
```diff
@@ -0,0 +1,20 @@
+#ifndef INTEGRATOR_SIZE_H
+#define INTEGRATOR_SIZE_H
```
maxpkatz, line 2: Can you also update the header guard

zingale: done

maxpkatz (approved):

PR #901, opened 2022-03-11
regenerate pynucastro nets to take advantage of recent optimizations

zingale:
note: subch2 will be updated separately


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2022-03-11-001/index.html

most of the test_rhs tests are roundoff, aside from subch, which has some Jacobian terms which are O(1.e-5) -- I will look at this to see if I understand what is happening there.

zingale:
``sn160`` doesn't compile because it has some terms like ``0.0e0_rt_rt`` in the ydot(H1) RHS

zingale:
there was a bug in the temperature derivative of 3-alpha -- it is fixed here:
https://github.com/pynucastro/pynucastro/pull/214/files

zingale:
this looks good now:
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2022-03-11-002/index.html

maxpkatz (approved):

PR #913, opened 2022-03-22
add a rate_low_cutoff_temp parameter

zingale:
this is used when integrating the system to zero out the rates
if the temperature drops below a certain value.  By default this
is set to -1.0, disabling the cutoff.  In the future, we can have
networks override this as needed, depending on how their rates
are constructed.

This works for both simplified-SDC and Strang.  For simplified SDC,
the advection terms are still included if we fall below the rate threshold.

This PR also removes the old NETWORK_HAS_CXX_IMPLEMENTATION
defines throughout (but leaves a single one in the main make include
for backwards compatibility)

### DIFF
--- a/Make.Microphysics_extern
+++ b/Make.Microphysics_extern
@@ -1,6 +1,10 @@
 # This is the main include makefile for applications that want to use Microphysics
 # You should set NETWORK_OUTPUT_PATH before including this file
 
+# for backwards compatibility for any application code that still looks for this
+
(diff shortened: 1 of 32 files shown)


zingale:
tests look good: 
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2022-03-22-002/index.html

zingale:
we also need to do this in `numerical_jacobian.H` since that calls actual_rhs() directly!

zingale:
I don't think that this is needed -- we should just set the burn cutoff in the application code.

PR #915, opened 2022-03-25
a slightly better subch2 network

zingale:
this adds Na23 + explicitly pulls in 2 rate sequences for
C12+O16 and O16+O16

This also uses symmetric screening and explicltly pairs
forward and reverse rates


zingale:
Here's the network plot

subch2

zingale:
I took out the C14 sequence

subch2

zingale:
okay updated once more... I added back in the missing Na23 and modified the neutron rates to skip the neutron capture.

subch2

maxpkatz (approved):

PR #916, opened 2022-03-28
sync simplified SDC clean_state up with Strang

zingale:


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2022-03-28-001/index.html

I think this is a good idea, just to sync up with Strang

maxpkatz (approved):

PR #917, opened 2022-03-31
clean up the simplified SDC numerical Jacobian a bit

### LINKED ISSUE #897: eliminate temporary ydot_react in vode_rhs_simplified_sdc.H
once we swap the order of the state to put (rho X) before (rho e), we can just directly fill ydot with the RHS and then fix it up to be the proper form for SDC.  Same (probably) for the Jacobian.  This will save on a lot of copying and memory.

To be safe, we can add some asserts that check that the index assumptions match.

zingale:
this removes an unneeded EOS call as well as eliminates a copy
(this will close #897)

### DIFF
--- a/integration/VODE/vode_rhs_simplified_sdc.H
+++ b/integration/VODE/vode_rhs_simplified_sdc.H
@@ -114,14 +114,22 @@ void jac (burn_t& state, dvode_t& vode_state, MatrixType& pd)
 
     }
 
-    // Call the specific network routine to get the Jacobian.
-
-    JacNetArray2D jac_react;
-
     if (jacobian == 3) {
         jac_info_t jac_info;
         jac_info.h = vode_state.H;
-        numerical_jac(state, jac_info, jac_react);
+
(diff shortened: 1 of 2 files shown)


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2022-03-31-004/massive_star-sdc.html

zingale:
it might be the resets that are done in `eos()`?

zingale:
all SDC tests pass: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2022-03-31-005/index.html

maxpkatz (approved):

zingale:
here's the Microphysics tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2022-07-28-001/index.html

zingale:
and the Castro tests:

http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2022-07-28-001/index.html

it looks good (that one failure is preexisting)

PR #934, opened 2022-07-03
hardcode the reaclib rates in the networks

zingale:
we no longer need the reaclib_rates_metadata.dat file


zingale:
here's a Castro run of the one test with a pynucastro net:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2022-07-03-002/subchandra.html

zingale:
tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2022-07-04-006/index.html

maxpkatz (approved):

PR #935, opened 2022-07-04
test_rhs only makes sense with jacobian = 1

zingale:
the idea of this test is to directly call the network's RHS functions,
so we only want to test the Jacobian that the network implements


zingale:
note: this will break one of the Microphysics test suite runs that explicitly set jacobian=2

maxpkatz (approved):

PR #944, opened 2022-07-11
make the Jacobian type part of dvode_type

zingale:
also switch some ints to short to save memory


zingale:
CPU tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2022-07-12-001/index.html

zingale:
I can confirm this compiles on crusher with ROCm 5.1

maxpkatz (approved):

PR #949, opened 2022-07-15
Template out T derivatives from screening methods

yut23:
also remove the commented-out density derivatives in screen5


zingale:
are roundoff diffs expected?

zingale:
oh, I see, there was a change from `X / T` to `X * (1.0 / T)`, so that explains the roundoff

zingale (approved):

PR #950, opened 2022-07-16
start templating pynucastro nets

zingale:


zingale:
tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2022-08-09-001/index.html

zingale:
looks like it is 10% faster on CPU with this change:
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/test_react_C-subch_approx-VODE-timings.html

PR #954, opened 2022-07-26
[WIP] Switch Jacobian type on a previous nonlinear solve failure

zingale:


zingale:
the idea behind this PR is that if the numerical Jacobian is not working well, we should attempt the analytic Jacobian.  This seems to work well for the Castro subchandra problem.

zingale:
I don't think this is needed anymore.  It seems like the main reason we have failures is due to Jacobian caching, so when the burn is hard, it's best to disable Jacobian caching.

PR #955, opened 2022-07-28
Simple sdc analytic jac memory

zingale:
this is similar to the change we did for the jacobian = 3 version in #917


zingale:
note: I've remove the (likely broken) code for the enthalpy-based Jacobian.  If and when we want to get MAESTROeX + simplified-SDC working again, we'll have to reimplement that (but it can be done much more cleanly).

maxpkatz (approved):

Issue #729, opened 2021-07-29
analytic Jacobian species derivatives are missing terms


zingale:
We want to express our Jacobian in terms of (rho, e, X_k), but the reaction system is more naturally in terms of (ρ, T, X_k).

Consider dX/dt = f(ρ, T, X_k) = f(ρ, T(ρ, e, X_k), X_k)

then ∂(dX/dt)/∂(X_k) = ∂f/∂(X_k) |_ρ,T + ∂f/∂T |_ρ,X_k ∂T/∂(X_k) |_ρ,e

we are missing the last term

zingale:
note: this is for the analytic Jacobian

PR #956, opened 2022-07-28
correct the species derivatives in the analytic SDC Jacobian

zingale:
This addresses #729 for simplified SDC.  It captures the (small) effect that the df/dX |_e means that if we hold e constant and change species, then T needs to change (this is all with rho not changing)


zingale:
depends on #955

zingale:
note that both `jacobian=2` and `jacobian=3` already correctly include these terms.

zingale:
here's the SDC test in the Microphysics test suite:
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2022-08-03-001/test_sdc_aprox13_C.html

maxpkatz (approved):

PR #957, opened 2022-08-02
fix the screening of derived ReacLib rates

zingale:
in some cases, the forward rates were derived from the reverse rates
via detailed balance in ReacLib, but we were screening symmetrically
based on the reverse rate (because ReacLib calls all derived rates
"reverse", even if they are exothermic)


aisclark91 (approved):
Reaclib, foward and reverse rate classification by Q, implies this correction.

PR #960, opened 2022-08-06
Nse solver

zhichen3:
This pr implements nse solver using newton-raphson method. 

I also copied burn_cell inside unit_test into test_nse with some modifications for testing nse solver .

Some network that I tested are aprox19 aprox21 subch2 and sn160, using input files provided in burn_cell. Some network like iso7 and aprox13 will never work, since their nuclei (nuc.N == nuc.Z, always) will always result in singular jacobian.


zhichen3:
hmm.. there seems to be typo in Make.package. I used to to compile with `NSE=TRUE`, but I later changed it to `USE_ACTUAL_NSE=TRUE` for using nse solver, and forgot to change it inside Make.package. 
But I'm not sure if it has effects on the tests...

zingale:
can you merge the current `development` into this PR?

zhichen3:
Just merged it!

zhichen3:
close this pr, same pr in #963

PR #961, opened 2022-08-08
Astrochem quokka

psharda:
Hi Mike, this is a draft pull request for merging my astrochem-quokka branch with the main Microphysics branch.


integration/VODE/actual_integrator.H
```diff
@@ -80,14 +80,16 @@ void actual_integrator (burn_t& state, Real dt)
 
 #endif
 
-    // Subtract off the initial energy (the application codes expect
-    // to get back only the generated energy during the burn).
+    if (!use_number_densities) {
+        // Subtract off the initial energy (the application codes expect
+        // to get back only the generated energy during the burn).
```
BenWibking, line 85: Since it's not _directly_ related to integrating number densities, it might be better to create a separate option for this. It's conceivable some application might want to enable this, but still integrate some other network that uses mass fractions, or vice verse.

unit_test/burn_cell_primordial_chem/gpuvolta.pbs
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
BenWibking, line 1: Probably you didn't mean to include this PBS script?

Make.Microphysics
```diff
@@ -1,5 +1,5 @@
-AMREX_HOME ?= /path/to/amrex
-MICROPHYSICS_HOME ?= /path/to/Castro
+AMREX_HOME ?= /scratch/jh2/ps3459/amrex
+MICROPHYSICS_HOME ?= /scratch/jh2/ps3459/Microphysics
```
BenWibking, line 2: I think you accidentally included this in the PR. Probably don't want to change the path for everyone.

BenWibking (commented):
A few comments. It looks like there were two files committed by accident, maybe? My only substantive comment is that it might make sense to separate the option to integrate number densities `use_number_densities` from the option that controls whether the initial specific internal energy is subtracted from the state.

PR #962, opened 2022-08-10
Implemented integration of primordial ISM chemistry networks

psharda:
I have implemented a primordial ISM chemical network in Microphysics that can integrate the chemistry required for Population III star formation MHD simulations. Following are the key additions/changes to the existing routines: 

1. EOS - added a sub-directory for the EOS of primordial ISM chemistry, based on EOS 'multigamma'. Updated the README file.
2. networks - added a sub-directory for the primordial ISM chemistry network that includes the actual_rhs.H file created using pynucastro.
3. integration - added two flags in the integrator which check for a.) if we are working with number densities, and b.) if we want the internal energy subtraction to be switched off. The first flag disables normalization (since we are not working with mass fractions, but number densities, for primordial ISM chemistry) and does not apply 'vode_failure_tolerance' as a ceiling. The second flag disables subtraction of the internal energy. The default values of these flags is set to false, such that other parts of Microphysics not using primordial ISM chemistry would continue to work as before.
4. unit_test - I have added a new burn_cell_primordial_chem unit test that runs the one-zone freefall collapse of a primordial molecular cloud and evolves the temperature and the abundances of different chemical species.

I have also added printing the internal energy for burn_state in interfaces/burn_type.H, and printing the jacobian terms in integration/VODE/vode_type.H

Please let me know your comments on this pull request.

Regards,
Piyush


zingale:
this is a large PR, so it will take some time to test and review.  It might be easier if it could be broken up into several smaller PRs.  For instance, a separate PR that just addresses the VODE bits?

zingale:
here are the tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2022-08-10-001/index.html

there are some crashes and lots of differences.  We should understand these.

psharda:
Thanks for reviewing Mike! I looked at the tests that failed to compile. It turns out the reason they were not compiling were because I was 17 commits behind the main. Once I imported these commits, I could compile these tests successfully. Can you please re-run the tests again and see if they pass this time? I am not sure how to run all these tests myself other than compile and run each individual one on my branch.

zingale:
updated tests:
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/

zingale:
There are 2 reasons for splitting:

1. this is a lot to review
2. we "squash on merge", so all of this will be put into a single commit.  So it's nice to have that single commit affect only one of the modules in Microphysics or do only one thing, since it helps debug later issues

So my suggestion is to try to break this into the following separate commits:

1. just the primordial_chem EOS
2. the integrator changes
3. the network + unit test

I think that if you do that, in that order, then they will build on one another and we can test each separately, and by the end, test the entire unit test.

Also, separately, we should update the docs.  In particular, I am curious about the advantage of integrating number densities instead of mass fractions.

PR #964, opened 2022-08-18
an even simpler subch net

zingale:
This is a further approximation of the subch_full network.  In
addition to removing the same 6 nuclei as subch_approx (by doing an
(a,p)(p,g) approximation), it also removes

* The reverse of C12+C12, C12+O16, and O16+O16 -- these reverse rates
  are not in aprox13

* The C12+Ne20 rates and their reverses

* (a,g) links between Na23 and Al27 and between Al27 and P31 -- these
  are not in aprox13


zingale:
here's what the network looks like:

subch_simple

networks/subch_simple/subch_simple.py
```diff
@@ -0,0 +1,133 @@
...
+
+    return subch
+
+def doit():
+
+    subch = get_library()
+
+    # these are the rates that we are going to allow to be optionally
+    # zeroed
+    r1 = subch.get_rate("p_c12__n13")
```
aisclark91, line 41: why `c12 + p -> n13` and `n13+a -> p + o16` are disabled but not removed?

zingale:
disabled means that the C++ network will produce runtime parameters that allow us to turn these off at runtime.  They are still part of the network though.

networks/subch_simple/subch_simple.py
```diff
@@ -0,0 +1,133 @@
...
+
+    # finally, the aprox nets don't include the reverse rates for
+    # C12+C12, C12+O16, and O16+O16, so remove those
+
+    rates_to_remove = []
+    for r in net.rates:
+        if sorted(r.products) in [[pyna.Nucleus("c12"), pyna.Nucleus("c12")],
+                                  [pyna.Nucleus("c12"), pyna.Nucleus("o16")],
+                                  [pyna.Nucleus("o16"), pyna.Nucleus("o16")]]:
+            rates_to_remove.append(r)
```
aisclark91, line 56: shouldn't be `r.reactants` instead of `r.products` ?

zingale: no, I am seeking to remove only the reverse of these rates.

zingale:
e.g., I can do

```
network.disable_p_c12__n13  = 1
```

in my `inputs` file to turn that rate off

networks/subch_simple/subch_simple.py
aisclark91: ahhh ok, makes sense!

aisclark91 (approved):

PR #967, opened 2022-08-19
have test_react output the burn state that took the longest

zingale:
this creates a file called zone_state.out with the burn_t

### DIFF
--- a/unit_test/test_react/main.cpp
+++ b/unit_test/test_react/main.cpp
@@ -193,44 +193,28 @@ void main_main ()
     AsyncArray<int> aa_num_failed(&num_failed, 1);
     int* num_failed_d = aa_num_failed.data();
 
+    BL_PROFILE("do_react");
+
+    // Do the reactions
+    auto const& ma = state.arrays();
+    auto const& ia = integrator_n_rhs.arrays();
+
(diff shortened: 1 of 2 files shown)


zingale:
I removed the OpenMP bits from here since I don't think we've been using them (especially on the GPU) and I wasn't sure how to do it together with this reduce

zingale:
oh, it looks like the openmp stuff is implicit here, in the `reduce_op.eval()`

zingale:
CPU tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/

zingale:
curiously this changes the test_nova integration on GPUs?
http://groot.astro.sunysb.edu/Microphysics/test-suite/gpu/2022-08-19-001/index.html

unit_test/test_react/main.cpp
```diff
@@ -191,44 +191,28 @@ void main_main ()
     AsyncArray<int> aa_num_failed(&num_failed, 1);
     int* num_failed_d = aa_num_failed.data();
 
-    {
```
maxpkatz, line 194: This curly brace pair ensures that the I/O at the end doesn't get counted in the `do_react` time.

zingale: fixed

maxpkatz (approved):

PR #969, opened 2022-08-22
add retry framework to burn

zingale:
This adds the ability to retry the burn in a zone if it fails, without having
to return to the application code.

It provides a few new parameters:

* use_burn_retry
* retry_swap_jacobians  (this will use the other Jacobian type for the retried burn)
* retry_rtol_spec, retry_rtol_enuc (the relative tolerance to use on the retry)
* retry_atol_spec, retry_atol_enuc (the absolute tolerance to use on the retry)

This works well near NSE where sometimes the other Jacobian type can get past
a stubborn burn.

The cost of this change is that we need to store an old copy of the burn_t, but
this is templated out if we don't run with use_burn_retry=1


zingale:
this seems to work, although it comes with the overhead of a backup `burn_t`

zingale:
CPU tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2022-09-07-008/index.html

integration/BackwardEuler/actual_integrator.H
```diff
@@ -6,22 +6,34 @@
...
+        be.atol_spec = retry_atol_spec; // mass fractions
+        be.atol_enuc = retry_atol_enuc; // energy generated
+
+        be.rtol_spec = retry_rtol_spec; // mass fractions
+        be.rtol_enuc = retry_rtol_enuc; // energy generated
+    }
 
     // set the Jacobian type
-    be.jacobian_type = jacobian;
+    if (!(is_retry && retry_swap_jacobian)) {
```
zhichen3, line 32: if both of them are false, then wouldn't it still swap the jacobian?

zingale: do you mean is `!(false && false) == true`?

zingale: indeed, I think we need to rewrite this logic

zingale: wait... I think it is okay.

This will evaluate to true is either or both of `is_retry` and `retry_swap_jacobian` are false, in which case it does not swap the jacobian.

only if both `is_retry = true` and `retry_swap_jacobian = true` does it do:
```
be.jacobian_type = (jacobian == 1) ? 2 : 1;
```
which is the swap.

right?

yut23: I did a quick truth table in a spreadsheet, and the logic is correct right now. It's definitely not obvious at first glance, though. I think swapping the if/else branches would help.

zingale: agreed.  It wasn't obvious to me either, and I wrote it ;)

zingale: updated

zhichen3 (approved):

PR #970, opened 2022-08-24
add the number of steps taken to the burn_t

zingale:
this is a better metric than the number of RHS calls when comparing different Jacobians.

test_react now outputs a summary as well.  Note that the sum() we were doing there was local, so it would not have worked with MPI.  This is now fixed.


zingale:
well... I guess it didn't matter with MPI anyway, since we only have 1 box...

zingale:
all tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2022-08-24-001/index.html

maxpkatz (approved):

PR #973, opened 2022-08-26
some more NSE cleaning

zingale:
this allows us to come in with a y_e and not rederive it from the
mass fractions.  In this manner, we can just use rho, T, and y_e as
inputs.  It also marks state as const in a few places.

### DIFF
--- a/nse/actual_nse.H
+++ b/nse/actual_nse.H
@@ -20,7 +20,7 @@ struct Newton_inputs{
 };
 
 template <typename T>
-T get_nse_state(T& state)
+T get_nse_state(const T& state)
 {
   // This function finds the nse state given the burn state or eos state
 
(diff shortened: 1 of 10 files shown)


nse/actual_nse.H
```diff
@@ -135,14 +134,16 @@ void nse_nr_solver(T& state, amrex::Real eps=1.0e-3_rt, int nloops=500){
...
   Newton_inputs f = nse_constraint(state);                    // get constraint eqs and jacobian
+
   amrex::Real det;                                            // store determinant for finding inverse jac
   decltype(f.jac) inverse_jac;                                // store inverse jacobian
   amrex::Real d_mu_p = 0.0_rt;                                // difference in chemical potential of proton
   amrex::Real d_mu_n = 0.0_rt;                                // difference in chemical potential of neutron
 
   // begin newton-raphson
-  for (int i = 0; i < nloops; ++i){
+  for (int i = 0; i < max_nse_iters; ++i){
```
zhichen3, line 146: is max_nse_iters declared somewhere?

zingale: I forgot to add the `_parameter` file -- I just pushed it

zhichen3 (approved):

PR #976, opened 2022-08-29
add a C++ version of Powell's hybrid method

zingale:
this is a conversion of the Fortran MINPACK hybrj to C++ as a
templated header-only library.  The main functional change is that
we assume square matrices now and we can pass an arbitrary payload
to the function and Jacobian.

This also hooks this solver into the new NSE solver.


zhichen3:
should we have some sort of check to test convergence?

zingale:
It has a built in convergence check

zhichen3:
When I give it a very wrong guess and the resulting mass fraction goes to inf, it still prints out stuff just fine. So I'm wondering if there should be any warnings about it.

zingale:
ah, yes, we should then.  Can you share an example of it going bad?

zhichen3:
I was testing subch_full (subch2 if I remember correctly) network and if I recall correctly, NR-solver had no problem solving it with my preset input file for subch2 (input_subch2). But when I tried to run that test now, it is having a hard time solving it. And the same goes for the hybrid solver where the output mass fractions clearly don't sum up to 1.

nse/actual_nse.H
```diff
@@ -80,86 +87,206 @@ T get_nse_state(const T& state)
...
+    hj.xtol = eps;
+    hj.mode = 1;
+
+    for (int j = 1; j <= 2; ++j) {
+        hj.diag(j) = 1.0_rt;
+    }
+
+    hybrj(hj, state, fcn_hybrid<T>, jcn_hybrid<T>);
+
+    if (!hj.info == 1) {
```
zhichen3, line 279: do you mean `hj.info != 1` ?. I was also testing it and it seems like hj.info is always 1 no matter if it converges to diverges.

zhichen3 (approved):

nse/actual_nse.H
zingale: I think I fixed this

zingale: oh I didn't.  I fixed it somewherre else.  I'll update this

zhichen3: yup. I saw it.

Issue #972, opened 2022-08-26
clean up actual_nse.H


zingale:
For the new NSE solver, we should create an `init_actual_nse()` that does the following:

* performs the `singular_network` check
* stores the index of the photoionization protons -- we cannot access `short_spec_names_cxx` on the GPU

This `init_actual_nse()` will be called once at initialize from the CPU.

PR #977, opened 2022-08-29
Nse init

zhichen3:
This pr attempts to address issue #972 .
Also gives a check on whether the input electron fraction is within the valid range. If not, the solver gets electron fraction by using the actual electron fraction.
Also renamed files and edited README since we are not doing any integration anymore, no more burning.

### DIFF
--- a/nse/actual_nse.H
+++ b/nse/actual_nse.H
@@ -11,26 +11,12 @@
 #include <AMReX_Array.H>
 #include <actual_network.H>
 #include <eos_composition.H>
-
-
-namespace NSE_INDEX
-{
-  extern bool initialized;
-  
-  // store indices:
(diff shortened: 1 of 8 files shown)


nse/actual_nse.H
```diff
@@ -11,7 +11,52 @@
...
 #include <actual_network.H>
 #include <eos_composition.H>
+#include <nse_index.H>
 
+// Initialization for nse, check whether network is valid and stores indices
+void init_actual_nse(){
+  
+  for (int n = 0; n < NumSpec; ++n){
+
+    // store index of photoionization proton by looking for Z=1 but exclude h1 which is always at index=0
```
zingale, line 21: I am not sure that this will always be true, but I think that it is reasonable for now

zingale (approved):

PR #982, opened 2022-09-17
improve and clean up nse solver

zhichen3:
Done:
1) Tried to update n_e while doing iterations, but it turns out it doesn't improve convergence and can make it worse. So I discarded that idea.
2) Since scipy.fsolve also returns incorrect solutions with no errors were raised,  I guess it is how hybrj is supposed to work.
3) Added convergence check for hybrj
4) I think mode 1 works better actually? So I changed it back to mode 1...
5) Transplant the fine-tuning of initial guess procedure in pynucastro for scipy.fsolve to hybrj solver, making it more robust at solving nse.
6) change criteria for making the table of solving initial guesses. With fine-tuning, it is easier to solve them.
7) modify nr procedure to use constraint equations for hybrj (although it is almost never a case we should use nr over hybrj, especially with the fine-tuning procedure added)

Need to do:
1) Add partition function and spin state to complete it.

### DIFF
--- a/nse/_parameters
+++ b/nse/_parameters
@@ -2,4 +2,5 @@
 
 max_nse_iters    integer    500
 
-use_hybrid_solver    integer   0
+use_hybrid_solver    integer   1
+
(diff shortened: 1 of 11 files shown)


zhichen3:
added partition function and spin to nse equation

zingale (approved):

PR #983, opened 2022-09-17
Adaptive Statistical Equilibrium

zhichen3:
Current task: Need to be able to pick out individual rates in the network.


zhichen3:
Seems like it is doing what I want it to do, but the test-reaction network doesn't pass the condition of the fast reaction network with extreme thermodynamic condition.

zhichen3:
Currently testing using subch_full nuclei. It seems like when I increased the density, then the temperature required for reaching NSE also increased. Especially when density is at ~1e9 g/cm^3, the temperature needs to be around ~10^10K.

 I also tried to use no screening and set u_c = 0, then it just failed to reach NSE when the density is beyond 1e8 g/cm^3.

I'll add some print statements to see where went wrong.

zingale:
I wonder if the issue is that we don't have the partition functions in yet.

zhichen3:
Oh, I thought we have it already because I saw there is a `partition_function.H` generated, but I guess its basically empty for now and only contains the spin. Yeah, then I guess that might be the issue.

zhichen3:
After several bug fixes, I think it now works. 

I had to set screening to null and set u_c=0, which makes sense since Alex explored these effects in pynucastro as well.  Surprisingly, chugunov2007 also worked okay with u_c turned on. Perhaps I can add the potekhin_1998 screening to microphysics that Alex implemented in pynucastro and see how it works. It should work well.

For the network, I just copied and pasted the `subch_simple.py` script to the `ase` network directory for testing, the only change is I added neutron to the network. 

For the unit_tests, I copied and pasted the `make_table` we had in `nse_solver` and added the part where we check whether we're in nse, perhaps I can just integrate them together, instead of creating a separate one in the test_ase.

Overall, it all just come down to calling in_nse(state, nse_state) by providing the current state and the nse state, and it returns a boolean stating whether we're in nse.

zhichen3:
the output from the test changed slightly when the order of the rates are changed (I updated the output from `subch_simple.py` using the latest pynucastro pr-494), Probably another bug somewhere.

zingale:
is this ready for review now?

zhichen3:
not yet, there is a small bug I'm fixing right now.

zingale:
just to understand -- there is nothing special about `networks/ase` except that you added neutrons to it, right?  This method should work for any pynucastro net that has neutrons?  (can it work with the other nets, like aprox13/19?)

zhichen3:
Right, I just added neutron to that network script. And it currently only work with pynucastro generated network as long as it has he4,proton,and neutron. It doesn't work with aprox networks for now because they have slightly different syntax in representing the rates, but it should not be difficult to implement them I think.

zingale:
can you add some documentation to https://amrex-astro.github.io/Microphysics/docs/dev/nse.html#self-consistent-nse explaining how to use this, including what requirements there are on networks.

zhichen3:
Yup, will do.

zingale:
tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2022-11-30-001/index.html

unit_test/test_ase/make_table/GNUmakefile
```diff
@@ -0,0 +1,45 @@
...
+
+# This sets the EOS directory in Castro/EOS -- note: gamma_law will not work,
+# you'll need to use gamma_law_general
+EOS_DIR     := helmholtz
+
+# This sets the network directory in Castro/Networks
+NETWORK_DIR := ase
+SCREEN_METHOD := null
+# SCREEN_METHOD := chugunov2007
+
```
zingale, line 29: does this currently not work with screening?

nse_solver/_parameters
```diff
@@ -1,6 +1,9 @@
 @namespace: nse
 
-max_nse_iters    integer    500
+max_nse_iters        integer    500
 
-use_hybrid_solver    integer   1
+use_hybrid_solver    integer    1
+
+cell_dx              real       1.0e6_rt
```
zingale, line 7: should this be the actual dx for the simulation? if so, we need to do this correctly, and the user should not have to set it.  Instead we should add `dx` to the `burn_t` and set it in Castro as we call the burner.

nse_solver/nse_solver.H
```diff
@@ -13,11 +13,147 @@
...
+	array_1d(j) = array_1d(j+1);
+	array_1d(j+1) = scratch;
+      }
+    }
+  }
+}
+
+
+void find_rate_indices(amrex::Array1D<int, 1, 3>& reactant_indices,
+		       amrex::Array1D<int, 1, 3>& product_indices, const std::string& rate_name){
```
zingale, line 41: I think that eventually we can just have pynucastro generate this for us

unit_test/test_ase/README.md
```diff
@@ -0,0 +1,7 @@
+# nse_example
```
zingale, line 1: this heading should be changed to `test_ase`

unit_test/test_ase/README.md
```diff
@@ -0,0 +1,7 @@
...
+
+`nse_example` finds the NSE mass fractions of a given state.
+The density, temperature, and composition are set in the
```
zingale, line 4: likewise, this should be updated to reflect what this example does

nse_solver/_parameters
zhichen3: Yes. I just put it there for reference for now.

zingale: okay, I did PR https://github.com/AMReX-Astro/Microphysics/pull/1074 to fix this.  So let's get that in and hooked into the application codes.

unit_test/test_ase/README.md
zhichen3: readme files are now updated.

zingale (approved):

PR #981, opened 2022-09-15
Added a new EOS folder for primordial chem

psharda:
Hi Mike,

I have managed to create a pull request specifically for the EOS part of primordial chemistry. Can you please let me know if this pull request seems okay? There are more changes down the line that I make to the EOS. I can generate another pull request for those later.

Regards,
Piyush

PR #985, opened 2022-09-23
Added more functionality to primordial chem EOS

psharda:
Hi Mike,

Thanks for approving pull request #981. Here are some more updates to the primordial_chem folder in the EOS. I will keep generating pull requests in small batches to make it easy to follow.

Regards,
Piyush


EOS/primordial_chem/actual_eos.H
```diff
@@ -202,41 +179,11 @@ void actual_eos (I input, T& state)
...
-            temp = state.p * state.abar * (m_nucleon / C::k_B) / dens;
-        }
-
-        break;
-
-    case eos_input_th:
-
-        // temperature, enthalpy and xmass are inputs
-#ifndef AMREX_USE_GPU
-        amrex::Error("eos_input_th is not supported");
```
zingale, line 235: I think that it would be good to have errors generated for each of the EOS input modes you don't support, just to prevent someone from using this EOS in a way that was not intended.

zingale:
can you add to the PR a description of the purpose of this change?  It is good practice for the PR to say why the change is being made, and since we squash on merge, we will retain that statement in the merged commit.

psharda:
Thanks for your comments, Mike. I have now included error messages for unsupported eos _types, and also updated the 'is_input_valid' function in actual_eos.H to reflect the same. The initial PR (#981) for EOS/primordial_chem simply introduced primordial_chem as an EOS folder with some basic functionality. The purpose of this PR is to build on #981 to properly implement a multigamma-type EOS for primordial chemistry given dens/temp or dens/eint as inputs (with number densities of all species).  This concludes all the changes that were required in the EOS for primordial_chem.

EOS/primordial_chem/_parameters
```diff
@@ -2,46 +2,60 @@
 
 eos_gamma_default       real          1.4
 
+=======
```
zingale, line 5: this looks like it is from a merge conflict? it should be removed

zingale:
tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2022-10-03-001/index.html

zingale (approved):

PR #997, opened 2022-10-03
Added the networks part of primordial ISM chemistry

psharda:
Hi Mike,

Now that the EOS part of primordial ISM chemistry is all done, I am generating this pull request to add its networks part. This pull request will create a new subfolder in Microphysics/networks, called primordial_chem, which will store the actual_rhs.H file (and associated files) for the primordial ISM chemistry network. Note that the actual_rhs.H file is quite big (has > 32000 lines of code, which includes the ydots and jacobians for 14 chemical species as well as the gas temperature). Please let me know if this is an issue, or if I should break this PR into smaller PRs.

Regards,
Piyush

PR #999, opened 2022-10-11
Changes to integration/VODE for primordial ISM chemistry

psharda:
Hi Mike,

This PR deals with changes I have made in VODE for primordial chemistry. All the changes basically reflect that since we use number densities for primordial chem, we want to disable some of the sanity checks in VODE that are appropriate for mass fractions. We also do not want to subtract the initial internal energy as is appropriate for nuclear astrochemistry. To disable the checks related to these, I have created two new flags in VODE/_parameters. The default values of these flags is such that users who do not use primordial chem will not need to worry about them, and the code should work just like before. Please let me know what you think.

PS: I am not sure why the changes to networks/primordial_chem are also being included in this PR, when we already approved and merged them in the PR yesterday (#997)


zingale:
GPU tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gpu/2022-11-18-002/index.html

I suspect that the one failure is due to optimization diffs

zingale:
CPU tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2022-11-18-001/index.html

zingale (approved):

PR #1000, opened 2022-10-12
we no longer need HAS_CXX_IMPLEMEMENTATION

zingale:
all nets are C++


zingale:
that CI failure is the timeout we've seen before with compiling primordial_chem

zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2022-10-12-001/index.html

maxpkatz (approved):

PR #1001, opened 2022-10-15
implement call_eos_in_rhs = 0 for simplified_sdc

zingale:
this basically reuses the T_fixed mechanism from drive_initial_convection


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2022-10-15/index.html

zingale:
this seems to make things worse, so it is quite possible that I have not implemented it correctly.

PR #1003, opened 2022-10-30
rename NSE_THERMO -> AUX_THERMO

zingale:
there is nothing about NSE in using the auxiliary state to define the composition, and our choice of the aux composition is sufficiently general enough that it should work for non-NSE applications.

This also renames set_nse_aux_from_X to set_aux_comp_from_X and puts it into eos_composition.H

Finally, we still define NSE_THERMO for backwards compatibility, but it will be removed once application codes are ported.


zingale:
Microphysics tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2022-10-30/index.html

zingale:
Castro needs to be updated in sync with this, so I'll do that now

zingale:
Castro tests pass:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2022-10-30/index.html

zhichen3:
I'm trying to review this. And but when I only add `USE_AUX_THERMO = TRUE` in the makefile, it seems that there is nothing generated in `AuxZero` in `network_properties.H`. The compiler is complaining because in `eos_composition.H` we're accessing it.

zingale:
Ah indeed.  Right now it would only work with aprox19.  I need to update that network script to always these aux quantities if we choose this option.

zingale:
okay, I think it should do it now, but I need to retest Microphysics

zingale:
tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2022-10-30-002/index.html

zhichen3 (approved):

networks/general_null/network_param_file.py
```diff
@@ -109,6 +111,27 @@ def parse(species, extra_species, aux_vars, net_file, defines):
 
         line = get_next_line(f)
 
+    # special case: if we are defining AUX_THERMO, then we want Ye,
+    # abar, and bea in the auxiliary variable list
+    if "-DAUX_THERMO" in defines:
+        index = get_object_index(aux_vars, "Ye")
+        if index <= 0:
```
yut23, line 118: These should check for `index < 0`, in case the variable is already at index 0

zingale: yes, of course.  Had my Fortran brain in :)

zingale: fixed in https://github.com/AMReX-Astro/Microphysics/pull/1012

PR #1005, opened 2022-10-30
rename actual_nse -> nse_net

zingale:
This is part of the clean up of the 2 different NSE approaches we have.

Depends on #1003


zhichen3 (commented):
still need to fix calling `init_actual_nse()` inside `nse_solver/make_table/main.cpp` . Also correct the error message inside `nse_solver.H` to call `init_nse_net()` when uninitialized, line 60. Perhaps also change the header guard in `nse_solver.H`.

zingale:
I think I fixed them

nse_solver/nse_solver.H
```diff
@@ -17,7 +17,7 @@
```
zhichen3, line 17: Do we want to change `_actual_nse_H` to something like `_nse_solver_H` ?

zingale: yeah, I'll change that

zingale: done

zhichen3 (approved):

PR #1006, opened 2022-10-30
aprox21 does not work with the NSE table

zingale:


zingale:
tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2022-10-30-001/index.html

zhichen3 (approved):

PR #1017, opened 2022-11-01
remove the USE_GPU_PRAGMA and our own CUDA settings

zingale:
this is from before AMReX had CUDA support


zingale:
GPU tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gpu/2022-11-01-001/index.html

maxpkatz (approved):

PR #1020, opened 2022-11-03
read the helmtable into local memory, bcast, and then copy to managed

zingale:
this gets around some MPI issues with GPUs


zingale:
GPU tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gpu/2022-11-03-001/index.html

zingale:
castro passes: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2022-11-03-001/index.html

zhichen3 (approved):

PR #1024, opened 2022-11-05
add an SDC github action

zingale:


.github/workflows/burn_cell_sdc.yml
```diff
@@ -0,0 +1,44 @@
...
+          echo $AMREX_HOME
+          if [[ -n "${AMREX_HOME}" ]]; then exit 1; fi
+          cd ../..
+
+      - name: Install dependencies
+        run: |
+          sudo apt-get update -y -qq
+          sudo apt-get -qq -y install curl cmake jq clang g++>=9.3.0
+
+      - name: Compile, burn_cell_sdc (subch_approx)
```
zhichen3, line 29: hmm, it says subch_approx but we're actually compiling aprox19?

zingale: good catch -- copy-paste error.  I've fixed it.

zhichen3 (approved):

PR #1026, opened 2022-11-12
Sync QSS integrator with new RHS scheme

maxpkatz:
This requires some minor changes to rhs.H so that the RHS infrastructure can separately return the forward and reverse terms, but it is worth it because we can remove most of the code duplication in this integrator. VODE should be unchanged.


networks/rhs.H
```diff
@@ -1420,13 +1449,25 @@ void rhs (burn_t& state, Array1D<Real, 1, neqs>& ydot)
 #endif
 
     // Compute the energy RHS term.
-    ydot(net_ienuc) = -sneut;
+    if constexpr (nrhs == 2 * neqs) {
+        ydot(2 * net_ienuc - 1) = 0.0;
+        ydot(2 * net_ienuc    ) = sneut;
```
zingale, line 1454: why is this positive here?

zingale: oh, is the idea that the second net_inuc are the negative terms?

zingale (approved):

networks/rhs.H
maxpkatz: Yes that's right. Same principle for the second slot on the species too. Just arbitrarily storing it as a positive number and then subtracting it later, but could have made the other choice.

zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2022-11-12-003/index.html
diffs are roundoff

PR #1037, opened 2022-11-15
remove the Strang self_heat parameter

zingale:
it is never used


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2022-11-15-002/index.html

PR #1042, opened 2022-11-16
get the BackwardEuler solver working with simplified-SDC

zingale:


zingale:
builds on #1041

zingale:
depends on #1043

zingale:
current status: simplified-SDC builds with `BackwardEuler`, but the `burn_cell_sdc` test fails with too many steps

zingale:
this works now

maxpkatz (approved):

PR #1043, opened 2022-11-16
move some general SDC routines out of VODE

zingale:
this allows other integrators to use them


zingale:
tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2022-11-16-002/index.html

maxpkatz (approved):

PR #1044, opened 2022-11-16
we don't need the first EOS call in the SDC burn

zingale:
it is overwritten when we do our first vode_to_burn


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2022-11-19-002/index.html

maxpkatz (approved):

PR #1045, opened 2022-11-17
more BackwardEuler refactoring

zingale:
this moves y and jac to the be_t
depends on #1041


zingale:
also depends on #1043

integration/BackwardEuler/be_integrator.H
```diff
@@ -328,7 +373,7 @@ int be_integrator (burn_t& state, be_t& be)
     if (ierr == BE_SUCCESS) {
         // do a final clean_state on the solution
 
-        clean_state(state);
+        //clean_state(state);
```
maxpkatz, line 376: What's up with this one?

zingale: oh, that's no longer necessary, since we do it in `actual_integrator.H`.  I'll remove it.

zingale: done

maxpkatz (approved):

PR #1046, opened 2022-11-17
template some VODE funcs to be used by other integrators

zingale:


zingale:
tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2022-11-19-001/index.html

maxpkatz (approved):

PR #1050, opened 2022-11-18
eliminate an eos_t temporary in the SDC code path with energy

zingale:


zingale:
tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2022-11-19/index.html

maxpkatz (approved):

PR #1060, opened 2022-11-22
template the VODE rhs/jac stuff

zingale:
this also lets BackwardEuler use these same functions


zingale:
I think that this is good for review now

maxpkatz (approved):

PR #1065, opened 2022-11-27
generalize the SDC helper routines

zingale:
this let's them be used by other integrators
also fix some includes
finally, we now explicitly pass time into the Jacobian wrapper


zingale:
this is ready for review

maxpkatz (approved):

PR #1066, opened 2022-11-27
some simplified-SDC cleaning

zingale:
rename fill_unevolved_variables to update_density_in_time -- this more accurately describes what it does.
also output the initial rhoe on failure


zingale:
note: this also simplifies the logic for determining the tolerances for the enthalpy-based integration (which we do not test :(

This is good to go now.

maxpkatz (approved):

Issue #1067, opened 2022-11-27
get rid of jacobian=3


zingale:
The idea behind `integrator.jacobian=3` was to do some parts numerically and some analytically, but I don't think that it really helps any.  We should remove that logic from VODE.  We will keep the implementation of the numerical Jacobian though, since other integrators can use that (like `BackwardEuler`).

PR #1069, opened 2022-11-28
remove jacobian = 3 option

zingale:
this did parts of the Jacobian numerically and the conversion to energy analytically.  But it never seems to have been a big win, and maintaining this code path is not worth it.


zingale:
closes #1067

zingale:
any application that used jacobian = 3 will now simply fall back to jacobian = 2, so no changes are needed.

maxpkatz (approved):

PR #1079, opened 2022-12-04
Add a safeguard to jac_term

maxpkatz:
This simply ensures that for d(f(Y_1)) / dY_2, Y_1 is actually present in a given RHS term. This is a redundant check for now which can give us some freedom later.


zingale:
tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2022-12-04/index.html

zingale (approved):

PR #1080, opened 2022-12-04
[WIP] redo the whole NSE update if we bail midway

zingale:
this allows to check if we are actually doing the remaining NSE correctly


zingale:
I don't think this is the right approach

PR #1081, opened 2022-12-04
Use rate containers in rhs_term and jac_term

maxpkatz:


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2022-12-04-001/index.html

zingale (approved):

PR #1086, opened 2022-12-08
mass fraction runtime parameter cleaning

zingale:
Lots of tests had the same code to interpret mass fraction runtime parameters.  This consolidates them all.

PR #1087, opened 2022-12-08
add aux to burn_cell so it works with aprox19 + NSE

zingale:
Also add a unit test (CI) that tests this out.


zingale:
depends of #1086

zhichen3 (approved):

PR #1089, opened 2022-12-10
update test_sdc_vode_rhs

zingale:
switch to the analytic Jacobian
also remove some copies between burn_t and eos_t


zingale:
fixes #1088

zhichen3 (approved):

PR #1090, opened 2022-12-10
[WIP[ Fix SDC+NSE update

zingale:
This removes the advective part from the energy generation rate, and as a result, should get the energy / T more accurately.


zingale:
closing this in favor of #1092

the two are largely the same, and I just need to pick one to go forward

PR #1096, opened 2022-12-15
start cleaning up the different NSE headers

zingale:
this is part of the way towards letting the approaches coexist


nse_solver/nse_check.H
```diff
@@ -0,0 +1,1091 @@
...
+#include <eos_type.H>
+#include <network.H>
+#include <burner.H>
+#include <extern_parameters.H>
+#include <cmath>
+#include <AMReX_Array.H>
+#include <actual_network.H>
+#include <eos_composition.H>
+#include <nse_index.H>
+
```
zhichen3, line 14: probably need to include `nse_solver.H` as well.

nse_solver/nse_check.H
```diff
@@ -0,0 +1,1091 @@
...
+
+template <typename T>
+bool in_nse(T& state){
+
+  // This function returns the boolean that tells whether we're in nse or not
+  // Note that it only works with pynucastro network for now.
+
+  // Get the nse state which is used to compare nse molar fractions.
+  const auto nse_state = get_nse_state(state);
```
zhichen3, line 971: I just realized I made a mistake here, it should use `get_actual_nse_state(state)`. Can you fix that for me?

zingale: fixed

nse_solver/nse_check.H
zingale: done

zhichen3 (approved):

PR #1098, opened 2022-12-16
for Strang + NSE, we need to call EOS

zingale:
we were not getting the internal energy consistent with the inputs before adding to the energy

### DIFF
--- a/integration/nse_update.H
+++ b/integration/nse_update.H
@@ -125,7 +125,7 @@ void sdc_nse_burn(burn_t& state, const Real dt) {
         dq = dq_out - rho_bea_old/rho_old;
 
         // convert the energy to erg / cm**3
-        enucdot  = dq * C::MeV2eV * C::ev2erg * C::n_A / dt;
+        enucdot = dq * C::MeV2eV * C::ev2erg * C::n_A / dt;
 
         // update the aux source terms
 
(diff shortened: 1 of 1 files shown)


zingale:
Here's a comparison, using the new inputs_256.test.

First `development` : 

nse_test_256_plt00228 development_slice

and now with this PR:

nse_test_256_plt00244_slice

Notice that on `development` the temperature drops considerably from the initial 5.e9 K down to ~ 1.e8 K.  This does not seem right (and still happens even if I zero out the energy generation rate in NSE).

zingale:
okay, something is wrong here, and the massive star looks very bad with this change.  Need to sort that out.

zingale:
okay, this PR is definitely needed.  The reason is that with Castro https://github.com/AMReX-Astro/Castro/pull/2310 we no longer subtract off the initial energy.  This means that when we exit NSE, we need to have the valid full energy.  However, since the `burn_t` comes in without the energy initialized, we need to call the EOS with the input rho, T, aux to get an initial valid e.

integration/nse_update.H
```diff
@@ -188,6 +188,10 @@ void nse_burn(burn_t& state, const Real dt) {
   Real dyedt;
   Real X[NumSpec];
 
+  // get the energy -- we are assuming that rho, T are valid on input
+
+  eos(eos_input_rt, state);
```
aisclark91, line 193: I wonder if it would be better to have eos_input_re in this routine, and limit the temperature below 1.0e9

zingale: e is not defined coming into this routine, so we need to use rho, T -- the idea is that we need to find the e to use here.

aisclark91 (approved):

PR #1110, opened 2023-01-13
in_nse() input change

zhichen3:
change self-consistent nse version of `in_nse()` to use`burn_t` only.


zingale:
we should also remove it from `networks/aprox19/nse_table_check.H`

zhichen3:
Ok, I removed the eos_t input version in `nse_table_check`

zingale (approved):

PR #1111, opened 2023-01-22
swtich Strang over to use T_fixed + update SDC + NSE

zingale:
we were not being consistent with how we handled T_fixed, which is set if we are doing drive_initial_convection (Castro Strang update coming soon).  The idea is that the zone's internal energy should not be computed off of T_fixed, but all of the reaction rates should be.  This update fixes some issues with NSE and Strang burning where were we computing e with T_fixed, thus changing the amount of internal energy in a zone.

### DIFF
--- a/nse_solver/nse_solver.H
+++ b/nse_solver/nse_solver.H
@@ -294,7 +294,7 @@ T get_nse_state(const T& state)
 
       nse_state.xn[n] = network::mion(n+1) * pf * spin / state.rho *
           std::pow(2.0 * M_PI * network::mion(n+1) *
-                   C::k_B * T_n / std::pow(C::hplanck, 2.0_rt), 3.0_rt/2.0_rt) *
+                   C::k_B * T_in / std::pow(C::hplanck, 2.0_rt), 3.0_rt/2.0_rt) *
           std::exp(exponent);
   }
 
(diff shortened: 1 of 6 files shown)


zingale:
This change is in support of https://github.com/AMReX-Astro/Castro/issues/2344

zingale:
regression tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2023-01-22-001/index.html

nse_solver/nse_solver.H
```diff
@@ -285,11 +290,11 @@ T get_nse_state(const T& state)
       Real exponent = amrex::min(500.0_rt,
                                  (zion[n] * state.mu_p + (aion[n] - zion[n]) * state.mu_n
                                   - u_c + network::bion(n+1)) /
-                                 C::k_B / state.T * C::Legacy::MeV2erg);
+                                 C::k_B / T_in * C::Legacy::MeV2erg);
 
       nse_state.xn[n] = network::mion(n+1) * pf * spin / state.rho *
```
aisclark91, line 295: the pf is dependent of the temperature as well, I wonder if its included?

aisclark91 (approved):

nse_solver/nse_solver.H
zingale: yes, that's accounted for, since we evaluate `tfactors` with the correct T, and then `pf` is computed from `tfactors`

PR #1113, opened 2023-01-28
some cleaning for the NSE table

zingale:
there is now a table_idx() function that can be used to get the correct index into the 1-d table.
this makes new constexpr for the table deltas
eliminates arrays to store T, rho, ye at the table points, since those are easily computed
replaces some magic numbers with more consts

### DIFF
--- a/networks/aprox19/actual_network.H
+++ b/networks/aprox19/actual_network.H
@@ -33,11 +33,12 @@ namespace network
 namespace table
 {
 
-  constexpr int npts = 46221;
 
   constexpr int ntemp = 71;
   constexpr int nden = 31;
   constexpr int nye = 21;
+
+  constexpr int npts = ntemp * nden * nye;
 
   constexpr Real dlogT = 0.02_rt;
   constexpr Real dlogrho = 0.10_rt;
--- a/networks/aprox19/actual_network.H
+++ b/networks/aprox19/actual_network.H
@@ -39,9 +39,9 @@ namespace table
   constexpr int nden = 31;
   constexpr int nye = 21;
 
-  extern AMREX_GPU_MANAGED amrex::Array1D<amrex::Real, 1, npts> ttlog;
-  extern AMREX_GPU_MANAGED amrex::Array1D<amrex::Real, 1, npts> ddlog;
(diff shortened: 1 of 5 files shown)


zingale:
this is ready for review.  It doesn't change any functionality, just makes things easier to understand / reuse

zingale:
tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2023-01-27/index.html

networks/aprox19/actual_network.H
```diff
@@ -33,15 +33,16 @@ namespace network
...
 
-  constexpr int npts = 46221;
 
   constexpr int ntemp = 71;
   constexpr int nden = 31;
   constexpr int nye = 21;
 
-  extern AMREX_GPU_MANAGED amrex::Array1D<amrex::Real, 1, npts> ttlog;
-  extern AMREX_GPU_MANAGED amrex::Array1D<amrex::Real, 1, npts> ddlog;
-  extern AMREX_GPU_MANAGED amrex::Array1D<amrex::Real, 1, npts> yetab;
```
yut23, line 44: These are also defined in `networks/aprox21/actual_network.H`. Should those be changed as well, or removed entirely?

networks/aprox19/actual_network.H
```diff
@@ -33,15 +33,16 @@ namespace network
 namespace table
 {
 
-  constexpr int npts = 46221;
 
   constexpr int ntemp = 71;
   constexpr int nden = 31;
   constexpr int nye = 21;
```
aisclark91, line 40: Moved global magic numbers into constexpr

sphinx_docs/source/nse.rst
```diff
@@ -178,6 +178,18 @@ We determine is a zone is in NSE according to:
 
 .. _self_consistent_nse:
 
+
+NSE table ranges
```
aisclark91, line 182: Writing an output of the table.

aisclark91 (approved):
Yes, indeed, now the NSE table is clearer with this PR.

zingale:
that namespace was removed only in `aprox21`, since the table is not meant to work with that net.

I think it is still okay in `aprox19`

PR #1114, opened 2023-01-28
implement tricubic interpolation for tabular NSE

zingale:


networks/aprox19/nse_table.H
```diff
@@ -98,96 +98,193 @@ int nse_get_ye_index(const Real ye) {
...
+    // f(x) = a (x - x_i)**3 + b (x - x_i)**2 + c (x - x_i) + d
+    // to the data (xs, fs)
+    // we take x_i to be x[1]
+
+    Real a = (3 * fs[1] - 3 * fs[2] + fs[3] - fs[0]) / (6 * std::pow(dx, 3));
+    Real b = (-2 * fs[1] + fs[2] + fs[0]) / (2 * dx * dx);
+    Real c = (-3 * fs[1] + 6 * fs[2] - fs[3] - 2 * fs[0]) / (6 * dx);
+    Real d = fs[1];
+
+    return a * std::pow(x - xs[1], 3) + b * std::pow(x - xs[1], 2) + c * (x - xs[1]) + d;
```
aisclark91, line 114: In the tricubic case, we push `cubic` along each direction, 16 in the ye direction, 4 in the T direction and 1 in the rho direction. I think the same philosophy should be done in the linear case, to write `linear` and push it 4 in the ye direction, then 2 in the T axis and finally 1 for rho.

zingale: I think that the linear is simple enough that we just fit a linear function of the form:
f(x,y,z) = A xyz + B xy + C xz + D yz + E x + F y + G z + H
to the 8 points and go with it.

aisclark91 (approved):

PR #1115, opened 2023-01-29
initial h1_index mass fraction in nse_solver

zingale:
the state was undefined and causing valgrind errors


aisclark91 (approved):
these execution errors can be a living nightmare!!

PR #1117, opened 2023-01-30
remove eta from tabular NSE

zingale:
we were not using this feature, and I don't think it's the correct approach


zhichen3:
you can also remove it in `nse_solver/_parameters`. I think I added it there to avoid compilation errors when doing nse_update before.

zingale:
> you can also remove it in `nse_solver/_parameters`. I think I added it there to avoid compilation errors when doing nse_update before.

done

zhichen3 (approved):

PR #1120, opened 2023-02-03
add the start of a new hot-CNO + 3-alpha + rp network

zingale:
this is a net that could be used for XRBs.  Not sure how high we need to go into the rp breakout


zingale:
closing in favor of #1145

PR #1122, opened 2023-02-05
hybrj and NSE routine fix.

zhichen3:
1) Fixed some bugs in the NSE routine.
2) Fixed uninitialized variable issues caused by minpack hybrid solver.
3) Fixed erroneous arithmetic operations during hybrid solver due to NaN, INF and division by zero.

### DIFF
--- a/nse_solver/nse_check.H
+++ b/nse_solver/nse_check.H
@@ -1029,6 +1029,9 @@ bool in_nse(burn_t& state){
   ener_gener_rate(ydot, enuc);
 
 #ifdef NEUTRINOS
+  // get abar and zbar
+  composition(state);
+  
   amrex::Real sneut, dsneutdt, dsneutdd, snuda, snudz;
   sneut5(state.T, state.rho, state.abar, state.zbar, sneut, dsneutdt, dsneutdd, snuda, snudz);
 #else
--- a/nse_solver/nse_check.H
+++ b/nse_solver/nse_check.H
@@ -654,14 +654,13 @@ void nse_union(const int nuc_ind_a, const int nuc_ind_b, amrex::Array1D<int, 1, 
 
 
 AMREX_GPU_HOST_DEVICE AMREX_INLINE
-void fill_pair_rate_index(int& pair_rate_index, const int& rate_index){
+int get_pair_rate_index(const int& rate_index){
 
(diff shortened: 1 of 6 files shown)


nse_solver/nse_check.H
```diff
@@ -684,7 +693,7 @@ void fill_pair_rate_index(int& pair_rate_index, const int& rate_index){
 
     if (found_reverse_rate){
       pair_rate_index = n;
-      break;
+      return;
```
zingale, line 696: since we don't care about the input value of `pair_rate_index`, maybe we should make this function explicitly return `pair_rate_index`, that way the interface is clearer

nse_solver/nse_solver.H
```diff
@@ -223,18 +223,20 @@ T get_nse_state(const T& state)
...
 
-  const amrex::Real A1 = -0.9052_rt;
-  const amrex::Real A2 = 0.6322_rt;
-  const amrex::Real A3 = -0.5_rt * std::sqrt(3.0_rt) - A1 / std::sqrt(A2);
 
   // Store nse_state
   T nse_state;
 
-  // Find n_e for original state;
+#if SCREEN_METHOD != 3
```
zingale, line 230: can you add a comment here explaining why only certain screening methods us this?

util/hybrj/hybrj_dogleg.H
```diff
@@ -100,10 +101,18 @@ void dogleg(amrex::Array1D<Real, 1, neqs * (neqs + 1) / 2>& r,
     for (int j = 1; j <= neqs; ++j) {
         Real temp = qtb(j);
         for (int i = j; i <= neqs; ++i) {
+
             wa1(i) += r(l) * temp;
             l++;
+	    if ( std::isnan(wa1(i)) || std::isinf(wa1(i)) ) {
```
zingale, line 107: do we know if there work on a GPU?

zingale: looks like maybe we want ``amrex::isnan()``

zhichen3: Ah okay. Didn't know `amrex::isnan()` exists.

zingale (approved):

PR #1133, opened 2023-02-14
modernize the subch_simple network script

zingale:
this also adds a hook for using partition functions to get reverse rates


zingale:
this is a cleaner version of #1002

aisclark91 (approved):

PR #1134, opened 2023-02-16
update the pynucastro nets to the latest pynucastro

zingale:
this will quiet compiler warnings


zingale:
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2023-02-16-002/index.html

I think the Urca test fails because it now incorporates the gamma photons
@biboyd ?

biboyd (approved):

PR #1138, opened 2023-03-02
add an option to ignore checking the timescale of the rate for NSE_NET

zhichen3:
add an option to ignore the check where we want the time scale of the rates to be smaller than the sound crossing time scale.

### DIFF
--- a/Make.Microphysics_extern
+++ b/Make.Microphysics_extern
@@ -136,6 +136,10 @@ ifeq ($(USE_NSE_NET), TRUE)
      DEFINES += -DNSE_NET
      DEFINES += -DNSE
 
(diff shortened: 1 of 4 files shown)


zingale:
is there a reason to compile this out?

zhichen3:
I think I changed it to a run-time option

zingale:
this also seems to have changed something in the fast rate section?

zingale (approved):

zhichen3:
oh, I uncommented one of the conditions in `find_fast_reaction_cycle`. That condition wasn't specifically mentioned in the paper but was sort of implied. So I commented it out before.

PR #1139, opened 2023-03-02
add another option to use nse mass fractions for NSE_NET check

zhichen3:
Since forward and reverse rates depend on the mass fractions, to do a more cruel estimation, we can use nse mass fractions instead. This way, the thermodynamic condition becomes the only deciding factor.

Note here we still need to make sure the current mass fractions are not terribly away from the nse mass fractions, so only allow it after the first check, which checks on whether the current mass fractions are close to the nse mass fractions.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,16 @@
+# 23.03
+
+  * updated all of the pynucastro networks to the latest
+    pynucastro version (#1134, #1136)
+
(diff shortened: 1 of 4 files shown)


zingale:
can you describe these options in the docs?

zingale (approved):

zingale:
there's a burn_cell failure

zhichen3:
just re-ran it and it passed xD

PR #1146, opened 2023-03-06
add another option to completely skip molar fraction nse check after integration fails

zhichen3:
here I add an option to completely skip the molar fraction check for self-consistent NSE only after the integration failed. With the previously added option, `nse_molar_independent`, we directly use the NSE mass fractions for comparing forward and reverse rates.
This way, we allow the system to burn for some time, and break out to NSE if it is having a hard time.

Also changed the input state for functions in nse_check.H to use `burn_t` only.


zhichen3:
With this feature added, I get something like this for the detonation problem:

image

zingale:
can you add this to the NSE docs?

zingale (approved):

PR #1153, opened 2023-03-15
address some more compiler issues

zingale:
this also changes the job_info test to test on an empty string with `string.empty()`


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2023-03-15-001/index.html

yut23:
Header guard in `nse_solver/nse_index.H` should be fixed too (I think that's the last one)

zingale:
fixed

PR #1156, opened 2023-03-16
pass tf_t as const reference to aprox rates

zingale:


yut23 (approved):

zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2023-03-16-002/index.html

Issue #1157, opened 2023-03-17
create an nse_interp interface that doesn't compute X


zingale:
For the NSE table, we often don't need the updated X's until we are done with an update and have the new state.  That means that we can benefit from having an interface that doesn't do the cubic interpolation on all 19 X's -- this should be a lot faster.

PR #1158, opened 2023-03-17
allow us to optionally skip interpolating X from the NSE table

zingale:
this can be expensive and unnecessary when we are iterating


zingale:
addresses #1157

yut23 (approved):

PR #1159, opened 2023-03-19
if the integrator said we entered NSE, then continue

zingale:
previously we would recompute in_nse() and use that, but sometimes the integrator is exploring a state and enters NSE while the check in the burner does not.  This happens even with the requirement that we integrator for 10 steps before following NSE.  Now we will always enter NSE if the integrator said we did


aisclark91 (approved):

zingale:
no longer sure if this is a good idea.  Maybe we need to bump up `MIN_NSE_BAILOUT_STEPS`

zingale:
I no longer think that this is a good idea

PR #1170, opened 2023-04-07
Optimized CMakeLists 

psharda:
Optimized CMakeLists and made them simpler. This optimization was also necessary to correctly define the scope of microphysics variables and compiler definitions when being used as an external submodule.


unit_test/CMakeLists.txt
```diff
@@ -1,35 +1,35 @@
...
 
 #add primordial_chem test
 add_subdirectory(burn_cell_primordial_chem)
 
-set(networkpropfile "${CMAKE_BINARY_DIR}/network_properties.H")
-set(networkfile "${CMAKE_CURRENT_SOURCE_DIR}/../networks/primordial_chem/pynucastro.net")
-set(networkdir "${CMAKE_CURRENT_SOURCE_DIR}/../networks/primordial_chem/")
-set(networkpropertiesfile "${CMAKE_CURRENT_SOURCE_DIR}/../networks/primordial_chem/NETWORK_PROPERTIES")
-set(networkheadertemplatefile "${CMAKE_CURRENT_SOURCE_DIR}/../networks/general_null/network_header.template")
+#set(networkpropfile "${CMAKE_BINARY_DIR}/network_properties.H")
```
zingale, line 7: do we need to keep all of this commented-out code? or can we just delete it?  It'll be in the git history

psharda:
Ah yes, I wanted to keep it in case we need it in the future, but I could just access it from the git history. I removed all the old code.

zingale (approved):

PR #1171, opened 2023-04-13
Added WORKING_DIRECTORY for CMake execute_process scripts

psharda:
This PR specifies the WORKING_DIRECTORY for execute_process commands invoked in root CMakeLists.txt. It is not necessarily the case that current directory = CMAKE_BINARY_DIR, so explicitly specifying the working directory is important.

PR Ready for review.


psharda:
Also, depending on whether unit_test is built or not, `write_probin.py` now has slightly different arguments. When not building the unit_test, we do not want to write unit_test params in `extern_parameters.H/cpp`

zingale (approved):

PR #1174, opened 2023-04-29
remove abort_on_failure

zingale:
it is now the responsibility of the application code to check the success flag in burn_t.  This is because we were not aborting when on GPUs, so it is better if the application code makes these decisions.


sphinx_docs/source/integrators.rst
```diff
@@ -75,7 +75,7 @@ The input is a ``burn_t``.
 .. note::
 
    For the thermodynamic state, only the density, temperature, and
-   mass fractions -- we explicitly compute the internal energy
+   mass fractions are used directly&mdash;we compute the internal energy
```
maxpkatz, line 78: looks like github doesn't like that character...

zingale: that's embedded HTML for the em-dash

maxpkatz (approved):

PR #1176, opened 2023-04-29
add codespell CI

zingale:


neutrinos/sneut5.H
```diff
@@ -285,7 +285,7 @@ void sneut5(const Real temp, const Real den,
     // theta is sin**2(theta_weinberg) = 0.2319 plus/minus 0.00005 (1996)
     // xnufam is the number of neutrino flavors = 3.02 plus/minus 0.005 (1998)
     // change theta and xnufam if need be, and the changes will automatically
-    // propagate through the routine. cv and ca are the vektor and axial currents.
```
maxpkatz, line 288: Hahaha wasn't there a reason we did this

maxpkatz (approved):

neutrinos/sneut5.H
zingale: oh yes, I think it was an openacc compilation bug?  the compiler died even if vector was in
the comments?

PR #1178, opened 2023-05-04
fix cuda compile error with nse_net

zhichen3:


zingale:
can you tell me the command you are using to build this?

zingale:
one issue is that I don't think any of the hybrj stuff is marked with AMREX_GPU_HOST_DEVICE AMREX_INLINE

zhichen3:
I think it's working. Whenever we use `hybrj`, we just need to create two functions named `jcn` and `fcn` in the header file. 
I can't test it right now since perlmutter is down.

zhichen3:
tested on groot, seems to work.

util/hybrj/hybrj.H
```diff
@@ -12,26 +12,34 @@
...
-           const T& data,
-           std::function<void(amrex::Array1D<Real, 1, neqs>&,
-                              amrex::Array1D<Real, 1, neqs>&,
-                              const T&, int&)> fcn,
-           std::function<void(amrex::Array1D<Real, 1, neqs>&,
-                              amrex::Array2D<Real, 1, neqs, 1, neqs>&,
-                              const T&, int&)> jcn) {
+           const T& data
+	   //,
+           // std::function<void(amrex::Array1D<Real, 1, neqs>&,
```
zingale, line 19: can you just remove these instead of commenting them out?

util/hybrj/hybrj.H
```diff
@@ -12,26 +12,34 @@
...
+           //                    amrex::Array2D<Real, 1, neqs, 1, neqs>&,
+           //                    const T&, int&)> jcn
+	   ) {
 
     // the purpose of hybrj is to find a zero of a system of
     // n nonlinear functions in n variables by a modification
     // of the powell hybrid method. the user must provide a
     // subroutine which calculates the functions and the jacobian.
     //
+    // Supposely we ask user to supply fcn and jcn, where
```
zingale, line 32: change this to say that we require fcn and jcn to be fined in the program unit where we include this header

util/hybrj/hybrj.H
```diff
@@ -12,26 +12,34 @@
...
     // fcn jcn are the names of the user-supplied subroutine which
     // calculates the functions and the jacobian.
     //
     // void fcn(Array1D<Real, 1, neqs>& x, Array1D<Real, 1, neqs>& fvec, int& iflag)
     // void jcn(Array1D<Real, 1, neqs>& x, Array2D<Real, 1, neqs, 1, neqs> fjac, int& iflag)
     //
+    // However, right now we run into issues when compiling GPU.
+    // So right now we use global function fcn and jcn to calculate
+    // the function and the jacobian.
+    //
```
zingale, line 42: all of this can be removed (starting with line 33)

zingale (approved):

PR #1182, opened 2023-05-10
add an explicit dependence for extern_parameters.cpp

zingale:


yut23:
Isn't this already on line 154?

Issue #1184, opened 2023-05-14
treat compiler warnings as errors


BenWibking:
There are a few compiler warnings that are causing the Quokka CI to fail, since we have now enabled treating warnings as errors in our builds (`-Werror -Wall -Wextra -Wno-unused-parameter -Wno-unknown-pragmas -Wno-strict-aliasing`).

Is this something that can be enabled for Microphysics builds?

zingale:
yeah, I think it's a good idea.

BenWibking:
@psharda can you add this to the CMake build? all it requires is:
```
add_compile_options(-Werror -Wall -Wextra -Wno-unused-parameter -Wno-unknown-pragmas -Wno-strict-aliasing)
```

zingale:
I think this is fixed now

PR #1187, opened 2023-05-15
cmake: compile with warnings as errors

BenWibking:
This sets the CMake build to also compile treating warning as errors.

Fixes https://github.com/AMReX-Astro/Microphysics/issues/1184.


networks/primordial_chem/actual_rhs.H
```diff
@@ -20,7 +20,7 @@ AMREX_GPU_HOST_DEVICE AMREX_INLINE
 void rhs_specie(const burn_t& state,
              Array1D<Real, 1, neqs>& ydot,
              const Array1D<Real, 0, NumSpec-1>& X,
-             Real const z) {
+             Real const /*z*/) {
```
BenWibking, line 23: This parameter is unused in this function, so we comment it out to avoid a warning.

zingale: you can also use `amrex::ignore_unused()`

BenWibking: ah, that's true. i'm not sure why we have this paramter at all here, though, to be honest. @psharda do you know why this is here?

zingale (approved):

PR #1188, opened 2023-05-18
implement the NSE+SDC update for the self-consistent / network NSE

zingale:


integration/nse_update_simplified_sdc.H
```diff
@@ -193,4 +175,170 @@ void sdc_nse_burn(burn_t& state, const Real dt) {
...
+
+    state.y[SRHO] += dt * state.ydot_a[SRHO];
+    state.y[SMX] += dt * state.ydot_a[SMX];
+    state.y[SMY] += dt * state.ydot_a[SMY];
+    state.y[SMZ] += dt * state.ydot_a[SMZ];
+
+    // predict the U^{n+1,*} state with only estimates of the X
+    // updates with advection to dt
+
+    burn_t burn_state;
```
zhichen3, line 238: we need to store mu_p and mu_n to the `burn_state` from `state` I think.

integration/nse_update_simplified_sdc.H
```diff
@@ -193,4 +175,170 @@ void sdc_nse_burn(burn_t& state, const Real dt) {
...
+    }
+
+
+    // density and momenta have already been updated
+
+    // update the total and internal energy now
+
+    state.y[SEINT] += dt * state.ydot_a[SEINT] + dt * rho_enucdot;
+    state.y[SEDEN] += dt * state.ydot_a[SEDEN] + dt * rho_enucdot;
+
```
zhichen3, line 340: also need to update mu_p and mu_n after the sdc_burn I think

zingale: yes, I think so, but that means that we need to include an `SMUP` and `SMUN` index in `burn_type.H` I think?

zhichen3: we could do that, but right now in `Castro_react.cpp` `reac_state()` for the simplified-sdc version, I just have it storing `burn_state.mu_p` and `burn_state.mu_n` so we can simply do `burn_state.mu_p = state.mu_p` here. 

If we want to include `SMUP` and `SMUN`, then we'll change them there too.

zingale: indeed, that's easier

integration/nse_update_simplified_sdc.H
```diff
@@ -193,4 +175,170 @@ void sdc_nse_burn(burn_t& state, const Real dt) {
...
+        // update (rho e)^{n+1} based on the new energy generation rate
+        rhoe_new = state.y[SEINT] + dt * state.ydot_a[SEINT] + dt * rho_enucdot;
+
+        // call the EOS to get the updated T*
+
+        Real T_new;
+        burn_state.rho = state.y[SRHO];
+        burn_state.e = rhoe_new / state.y[SRHO];
+        for (int n = 0; n < NumAux; n++) {
+            burn_state.xn[n] = rhoX_new[n] / state.y[SRHO];
```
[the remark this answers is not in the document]
zingale: updated

integration/nse_update_simplified_sdc.H
```diff
@@ -193,4 +175,170 @@ void sdc_nse_burn(burn_t& state, const Real dt) {
...
+        Real rho_bea_tilde = state.y[SRHO] * dq_out - dt * ydot_a_BA;
+        Real rho_dBEA = rho_bea_tilde - rho_bea_old;  // this is MeV / nucleon * g / cm**3
+
+        // convert the energy to erg / cm**3
+        rho_enucdot  = rho_dBEA * C::MeV2eV * C::ev2erg * C::n_A / dt;
+
+        // update the new state for the next pass
+
+        for (int n = 0; n < NumSpec; n++) {
+            rhoX_new[n] = nse_state.xn[n];
```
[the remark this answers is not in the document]
zingale: updated

integration/nse_update_simplified_sdc.H
zingale: updated

integration/nse_update_simplified_sdc.H
```diff
@@ -193,4 +175,179 @@ void sdc_nse_burn(burn_t& state, const Real dt) {
...
+        rhoe_new = state.y[SEINT] + dt * state.ydot_a[SEINT] + dt * rho_enucdot;
+
+        // call the EOS to get the updated T*
+
+        Real T_new;
+        burn_state.rho = state.y[SRHO];
+        burn_state.e = rhoe_new / state.y[SRHO];
+        for (int n = 0; n < NumSpec; n++) {
+            burn_state.xn[n] = rhoX_new[n] / state.y[SRHO];
+            state.y[SFS+n] = rhoX_new[n];
```
zhichen3, line 271: I think it should be `burn_state` not `state`.

zingale: yes. you are right

zingale: fixed

zhichen3 (approved):

PR #1191, opened 2023-05-21
add a Runge-Kutta-Chebyshev (RKC) integrator

zingale:
This is a port of the integrator from Sommeijer et al. 1997.  It is fully explicit, so no
Jacobian is needed, but it does need an estimate of the spectral radius, which is
computed internally.


zingale:
I've tested this on Frontier with HIP and it works well

aisclark91:
yes, I have reviewed carefully, and works.

PR #1194, opened 2023-05-22
fix primordial_chem CI

BenWibking:
This runs the burn_cell_primordial chem unit test and should fail if the test does not pass. Previously, we were only building the test, not actually running it.

*TODO*: we still need to make sure that `burn_cell_primordial_chem` returns a non-zero exit code from the process if VODE fails or the results don't match expectations.


BenWibking:
Moving this PR to a fork.

PR #1056, opened 2022-11-21
enforce namespaces for the runtime parameters

zingale:

PR #1196, opened 2023-05-22
Using namespaces in CMake

psharda:
Following from #1056 we now use namespaces in CMake build for Microphysics.

The primordial chem test passes for me both with and without CMake.


zingale (approved):

BenWibking (approved):
Well, that made it look easy. Approved!

PR #1198, opened 2023-05-25
Use redshift as a runtime parameter in primordial chem

psharda:
We have to assume a certain redshift for doing Pop III simulations with primordial chemistry. We use z = 30, but it is also common to use a value between z = 20-30. With this PR, the user can define the redshift they want as a runtime parameter.

@zingale - I don't think Microphysics already has a redshift parameter embedded somewhere, right?

### DIFF
--- a/networks/primordial_chem/_parameters
+++ b/networks/primordial_chem/_parameters
@@ -3,3 +3,5 @@
 
 # cutoff for species mass fractions
 small_x                              real               1.e-100
+# assumed redshift for primordial chem (Pop III star formation)
+pc_redshift                          real               30e0
--- a/networks/primordial_chem/_parameters
+++ b/networks/primordial_chem/_parameters
@@ -4,4 +4,4 @@
 # cutoff for species mass fractions
 small_x                              real               1.e-100
 # assumed redshift for primordial chem (Pop III star formation)
-pc_redshift                          real               30e0
+redshift                          real               30e0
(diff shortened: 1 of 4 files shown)


zingale:
for us, we do:
```
constexpr z_reshift = 0.0
```

it's highly accurate ;)

psharda:
Haha indeed!

So, do you think it is better to define a separate redshift parameter for primordial chem? This was my thinking, which is why I prefixed it with `pc_`

zingale:
yes, I think this is fine.

networks/primordial_chem/_parameters
```diff
@@ -3,3 +3,5 @@
 
 # cutoff for species mass fractions
 small_x                              real               1.e-100
+# assumed redshift for primordial chem (Pop III star formation)
+pc_redshift                          real               30e0
```
BenWibking, line 7: since it's very unlikely another network will have a redshift parameter, maybe just make it `redshift`?

psharda: Well, we will include more ISM chemistry networks in Microphysics which may have a redshift dependence. For example, the earlyUniverse network, which we would want to run if we started from cosmological initial conditions.

BenWibking: Isn't it ok to have a parameter with the same name for that network, though?

psharda:
ready for review @zingale @BenWibking

BenWibking (approved):

Issue #1212, opened 2023-06-09
burn_cell_primordial_chem should test if answer is correct


BenWibking:
We need to add a check that the answer for the `burn_cell_primordial_chem` test is right.

Since we're using CTest, this can take advantage of the existing CI tests by having a this function return 0 or 1 based on whether the species fractions and internal energy are correct to some tolerance:

https://github.com/AMReX-Astro/Microphysics/blob/bcf06921ae6144e0f1d24ceb8181af43357e9f84/unit_test/burn_cell_primordial_chem/main.cpp#L31

Then this can be changed to:
```
int status = burn_cell_c();
```
and then adding
```
return status;
```
at the end of `main()`.

PR #1214, opened 2023-06-12
Compare primordial chem CI output to reference solution

psharda:
In addition to building the primordial chem unit test, we also want to make sure that the comparison to reference solution passes.

The way we test this as follows: we compare a select number of lines from the test output to the stored output. These lines describe the final temperature, internal energy, density, and number densities of electrons, H, H2 and He. If the ratio of the test values to the reference values is larger than 2, the CI will return an exit code 1 and fail the github action.

Closes https://github.com/AMReX-Astro/Microphysics/issues/1212.


psharda:
@zingale the `mac_os_burn_cell_primordial_chem` test is failing because it cannot install some dependencies? I don't think my changes should cause this issue.

BenWibking:
> @zingale the `mac_os_burn_cell_primordial_chem` test is failing because it cannot install some dependencies? I don't think my changes should cause this issue.

No, that's a bug in Homebrew formula for python. It might be possible to workaround this temporarily by removing this line from `.github/workflows/macos_build_cell_primordial_chem.yml`: https://github.com/AMReX-Astro/Microphysics/actions/runs/5243608479/workflow?pr=1214#L14.

zingale:
no clue.  I don't touch Macs ;)

.github/workflows/burn_cell_primordial_chem.yml
```diff
@@ -30,5 +32,32 @@ jobs:
...
+              echo "Value in test.out: $value1"
+              echo "Value in reference_solution.out: $value2"
+              echo "Ratio: $ratio is more than allowed threshold of $threshold"
+              echo
+              error_found=true
+            fi
+          done
+          if [[ $error_found == true ]]; then
+            exit 1
+          fi
```
BenWibking, line 62: It might be good to put this in a self-contained shell script, and just run the shell script from this file instead.

.github/workflows/burn_cell_primordial_chem.yml
```diff
@@ -30,5 +32,32 @@ jobs:
...
           make -j 2
-          ./main1d.gnu.DEBUG.ex inputs_primordial_chem
+          ./main1d.gnu.DEBUG.ex inputs_primordial_chem > test.out
+
+      - name: Compare to stored output
+        run: |
+          set -e
+          cd unit_test/burn_cell_primordial_chem
+          line_numbers=(456 458 460 463 465 471 476)
+          threshold=2
```
BenWibking, line 42: A factor of 2 is quite a loose tolerance. I was thinking something closer to a few percent. Is that too strict of a tolerance?

psharda:
> This is the issue with Homebrew / GitHub Actions: [actions/setup-python#577](https://github.com/actions/setup-python/issues/577).
> 
> It's caused by GitHub's misconfiguration of the VM images used for running GitHub actions on macOS. The easy workaround is to avoid running `brew update` as part of the action.

Do you know which line in the .yml file is calling `brew update`? The one you referenced above is not there in the latest version of the .yml file

.github/workflows/burn_cell_primordial_chem.yml
psharda: I thought about it, but I don't know how the .yml files interact with the shell script. I just followed the syntax of other .yml files in the workflows.

BenWibking:
> > This is the issue with Homebrew / GitHub Actions: [actions/setup-python#577](https://github.com/actions/setup-python/issues/577).
> > It's caused by GitHub's misconfiguration of the VM images used for running GitHub actions on macOS. The easy workaround is to avoid running `brew update` as part of the action.
> 
> Do you know which line in the .yml file is calling `brew update`? The one you referenced above is not there in the latest version of the .yml file

On the current development branch it's still there: https://github.com/AMReX-Astro/Microphysics/blob/8c5b63da04424fb75978f7f1d1352d0e61783e46/.github/workflows/macos_build_cell_primordial_chem.yml#L14.

.github/workflows/burn_cell_primordial_chem.yml
BenWibking: Ok, that's fine. For reference, the commands in the .yml files should be the same as running things from the command line, so you should be able to run a shell script in the current directory with `./my_shell_script.sh`.

BenWibking:
It might be good to also make sure that the CI returns an error if VODE reports an internal error. When I tested on my local machine, I was able to trigger a VODE error by adjusting the tolerances, but the test still passed.

.github/workflows/burn_cell_primordial_chem.yml
BenWibking: @zingale do you have a recommendation on the tolerance?

zingale:
not really.  You know your system of equations better than I

psharda:
> It might be good to also make sure that the CI returns an error if VODE reports an internal error. When I tested on my local machine, I was able to trigger a VODE error by adjusting the tolerances, but the test still passed.

If VODE returns an error, the integration will stop and `test.out` will not be fully written. In such cases, now, `value1` will be 0, and comparison of 0 with the reference solution (`value2`) will lead to an error.

PR #1222, opened 2023-06-17
estimate the spectral radius via the Gershgorin circle theorem

zingale:


zingale:
an issue here is that we are integrating the system (X, e), but e is 10s of orders of magnitude larger than X, so it is not clear that the circle theorem works here, without a scaling

zingale:
this seems to work well now with RKC if we run with `integrator.scale_system = 1`

maxpkatz (approved):

PR #1223, opened 2023-06-17
Template VODE on number of integrator equations

maxpkatz:
With this change, the number of equations integrated by VODE is now counted from the components of the burn_t, rather than being hardcoded at compile time (at least for Strang; SDC will be handled in a future change). This sets up a possible future change where different burn state objects are passed to the integrator at different times. The motivating use case is Quokka's desire to integrate (xn, e) at some times and (e_gas, e_radiation) at other times. A future change could support the integration of an object other than burn_t, particularly an object that does not have xn but has (multigroup) radiation energy, as long as the integration tolerances (and any other auxiliary information) are made consistent.

### DIFF
--- a/integration/VODE/actual_integrator.H
+++ b/integration/VODE/actual_integrator.H
@@ -20,9 +20,9 @@ template <typename BurnT>
 AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
 void actual_integrator (BurnT& state, Real dt)
 {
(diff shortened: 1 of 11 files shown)


zingale:
can you add a comment about the motivation for this, so we can preserve that in the git history?

maxpkatz:
Done

zingale (approved):

PR #1231, opened 2023-06-19
make self-consistent nse compatible with non-neutron networks.

zhichen3:
Due to the missing neutron rates, it is difficult to group isotopes together into a single group. 

This pr mainly adds another logic of defining a `single_group` for non-neutron networks. If isotopes fail to form a single group and there is no neutron in the network, then use a second check where we consider nse if isotopes past Z=14(Si) that have odd N or even N form two distinct groups.

Most of the line changes are code formatting.


zhichen3:
image

This is detonation test problem result with subch_simple and USE_NSE_NET=TRUE

zhichen3:
One thing I forgot to mention is I regenerated subch_simple files with `DO_DERIVED_RATES=TRUE` and `symmetric_screening=False`. Need them to make mass-fractions in equilibrium match with mass-fractions from NSE equations. 

So I'm not sure if we want to do that for subch_simple or create an identical one but with different settings.

zingale:
that's a good point
I think that ultimately, we want to update the net to use those defaults.  Let's keep that on a branch until I finish the `subch_tols` paper, since I don't want to rerun those simulations which are almost done.

zingale (approved):

PR #1235, opened 2023-06-20
[WIP] template actual_integrator

zingale:
There is a lot of code duplication in VODE, BackwardEuler, and RKC, so this templating really cleans things up.


zingale:
going to do this in stages

PR #1236, opened 2023-06-20
Move cubic and quintic polynomial interpolations to util

maxpkatz:


zingale:
is there a particular motivation for this?

PR #1240, opened 2023-06-28
Use amu instead of 1/n_A for m_nucleon in gamma_law

psharda:
Is this what we'd want to do (for gamma_law)?


psharda:
> I'm fine with this, but I want to run it though our suites, since there will be a lot of small changes. But I think it is a good idea to sync up with the standard constants here.

Sounds good. We may also need to update the reference solution for the test that fails.

zingale:
Castro tests:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2023-07-03-001/index.html

BenWibking:
This looks good to me. @zingale do we need to update the reference solution for the `eos_cell` test? If so, how do we do that?

zingale:
sorry, I will do that.  I'll try to get this done today and do a PR to your PR

zingale:
There's a PR to @psharda 's PR, that once merged will fix the EOS test.  Then this is good to merge.

BenWibking:
> There's a PR to @psharda 's PR, that once merged will fix the EOS test. Then this is good to merge.

@zingale just wanted to check- did this PR get merged?

zingale:
@psharda said that you were trying to understand something before merging this
is it all good now?

BenWibking (approved):
Looks good to me.

BenWibking (approved):

BenWibking:
@zingale are you happy with this in its current form? It's ready to go on our end.

PR #1241, opened 2023-06-29
[WIP] implement scale_system for simplified-SDC

zingale:
in addition to scaling energy, we now also need to scale density, since we are integrating (rho X, rho e)


zingale:
one issue I need to be careful with is that for the Jacobian, when we do:  d(rho omegadot)/(d rho X) we are actually expressing this as domegadot/dX, so there is no density scaling.  But for the integrator, we are scaling by the initial density, rho_0, so perhaps we need to correct these terms with a (rho / rho_0) factor?

zingale:
closing in favor of the simpler #1253

PR #1243, opened 2023-06-29
simplify some logic in the SDC integrators

zingale:
basically, once we convert back from the integrator type to the burn_t, we should not have to use the integrator type any more.  This will make it easier to merge these code paths in the future.


zingale:
Castro SDC tests all pass:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2023-06-29-001/index.html

maxpkatz (approved):

PR #1265, opened 2023-07-07
Convert triple_alpha_plus_cago to new reactions scheme

maxpkatz:
This will change the rates for this network by about 1% due to a different implementation of the triple alpha and C12(a,g)O16 rates in this network compared to the aprox rates we use for the other networks.


networks/triple_alpha_plus_cago/actual_network.H
```diff
@@ -1,33 +1,222 @@
...
+namespace RHS {
+    
+    AMREX_GPU_HOST_DEVICE AMREX_INLINE
+    constexpr rhs_t rhs_data (int rate)
+    {
+        using namespace Species;
+        using namespace Rates;
+
+        rhs_t data{};
+
```
zingale, line 113: this should use the constructor now

unit_test/test_rhs/ci-benchmarks/triple_alpha_plus_cago.out
```diff
@@ -3,38 +3,38 @@
...
- J_helium-4_E                -9.2068265904e-12       2.9771882829e-13
- J_carbon-12_E               -9.2068599038e-12       7.2504009603e-13
- J_oxygen-16_E               -2.4841469758e-19       9.2068515755e-12
+ J_E_iron-56                                -0                     -0
+ J_helium-4_E                -7.8037009251e-12       2.9599266218e-13
+ J_carbon-12_E               -7.8037355427e-12       7.2503604192e-13
+ J_oxygen-16_E                3.1567282452e-41       7.8037268883e-12
  J_iron-56_E                                 0                      0
- J_E_E                           -691898.93303           63621503.552
+ J_E_E                            -679969.8743           53925576.227
```
zingale, line 40: it looks like the biggest diff is in J_E_E -- were we doing something wrong before?

maxpkatz: Yes. This line that increases the CF88 rate by 1.7 had a typo in the d/dT derivative of the rates, which then polluted d/de downstream of that.

maxpkatz: https://github.com/AMReX-Astro/Microphysics/blob/23.07/networks/triple_alpha_plus_cago/actual_rhs.H#L234

zingale (approved):

PR #1270, opened 2023-07-08
update adsurl for refs.bib

zhichen3:


zhichen3:
never mind, this only fixes adsurl link error.

zingale (approved):

PR #1277, opened 2023-07-15
New Ci-Benchmark for the burn_cell test

aisclark91:
Due to differences in the units conversion factors and the number of decimals between the existing tabular rates and the ones of the pynucastro PR #550, the existing benchmark will differ.


yut23:
I get the same output as Github when I run this locally (`final T = 6608892764`), so it looks like the benchmark output just needs to be updated.

zingale:
I get the same

zingale (approved):

PR #1287, opened 2023-07-21
Update pynucastro networks

yut23:
Regenerated from the current pynucastro `main` branch: https://github.com/pynucastro/pynucastro/commit/556584df5085b578ead191e73fcb483e467170e0.

Also add a python script to do this with a single command.


yut23:
Going to wait for pynucastro/pynucastro#580 to be merged

zingale:
do we want to wait until we do the PR that changes the interpolation for tables to work on logs?

yut23:
Not sure. What do the tests look like now?

zingale:
I'm not sure I understand.

We haven't implemented that yet.  So once we do implement the proper interpolation, we'd want to regenerate these again.

yut23:
Oh, I see. I was wondering if that would give bigger diffs on the regression tests. I think the Suzuki table updates are the only thing that would change the results in this PR right now.

zhichen3:
I was trying to build subchandra with CUDA on perlmutter with updated network, but it seems like it's complaining that `std::upper_bound` doesn't work with cuda.
```
error: calling a __host__ function("T1  ::std::upper_bound<const double *, double> (T1, T1, const T2 &)") from a __host__ __device__ function("part_fun::interpolate_pf<(int)72> ") is not allowed
```
I also tried on groot, getting the same error, but I do see the `cuda` test passed here...

yut23:
That's unfortunate, we can probably just write a manual binary search instead (or maybe reuse `vector_index_lu()` from the tabular rates). My guess for why the test passes is that `interpolate_pf()` is a templated function and so it doesn't get compiled unless it's used somewhere. The CUDA test doesn't build with any networks that use partition functions; maybe we should add that?

zhichen3:
we do compile test_react with subch_simple in the cuda test though?

yut23:
Yes, but `DO_DERIVED_RATES` is false. https://github.com/AMReX-Astro/Microphysics/blob/37551260a0c17a85dd6d08fc289a64c38dfa46a6/networks/subch_simple/partition_functions.H#L66-L81

zhichen3:
ah okay. We could build with`ase` network, it has `DO_DERIVED_RATES=True`.

yut23:
Yeah, that would be good since it exercises approximate rates too.

zingale:
with the release of pynucastro 2.1.0, I think we are ready for this now.  We should update to get the latest table changes.

zingale:
I think we need to update some benchmarks now

zingale (approved):

PR #1290, opened 2023-07-22
Remove unneeded actual_rhs.H from new network scheme

maxpkatz:


maxpkatz:
The Castro test is failing on this PR but that's because we have it only looking at the last release. The relevant change has already been incorporated in AMReX-Astro/Castro#2534.

zingale (approved):

PR #1297, opened 2023-07-24
Update CUDA compilation test to cover more cases

yut23:
This adds CUDA coverage for tabular rates, partition functions, and self-consistent NSE.


zingale:
to keep the test compilation time down, I would suggest replacing the current subch_simple with the ase network.  ase tests everything that is in subch_simple already, so we don't need them both

yut23:
In that case we might as well use URCA-simple instead of subch_simple, since that will cover tabular rates too

zingale (approved):

PR #1298, opened 2023-07-24
Rewrite dedT in gamma_law EOS

psharda:
Hi @zingale , is it okay if we rewrite `state.dedT` in this way? It is identical to the original expression, but doesn't directly use other `state` quantities. 

If we can make this change, it would make it easy to extract `dedT` from Microphysics for Quokka without needing to supply a temperature.


BenWibking:
For a constant-gamma EOS, this works. We still need to compute $de/dT$ for more general EOSes, though, so I'm not sure how useful this is.

psharda:
> For a constant-gamma EOS, this works. We still need to compute de/dT for more general EOSes, though, so I'm not sure how useful this is.

If we don't do this, then the call to get dedT from Microphysics in ComputeEintTempDerivative in EOS.hpp will return a NAN, because we don't give it a temperature as an input.

BenWibking:
> > For a constant-gamma EOS, this works. We still need to compute de/dT for more general EOSes, though, so I'm not sure how useful this is.
> 
> If we don't do this, then the call to get dedT from Microphysics in ComputeEintTempDerivative in EOS.hpp will return a NAN, because we don't give it a temperature as an input.

We could give it the temperature as an input, no? Or is this not possible to give it the temperature for some reason?

psharda:
> > > For a constant-gamma EOS, this works. We still need to compute de/dT for more general EOSes, though, so I'm not sure how useful this is.
> > 
> > 
> > If we don't do this, then the call to get dedT from Microphysics in ComputeEintTempDerivative in EOS.hpp will return a NAN, because we don't give it a temperature as an input.
> 
> We could give it the temperature as an input, no? Or is this not possible to give it the temperature for some reason?

Well, if we wanted to keep the current structure of this function as it is, we don't want to give temperature as an input. My aim is to not have to change the function arguments.

BenWibking:
I don't know what you're referring to. For the function in EOS.hpp, the temperature is already a function argument: https://github.com/quokka-astro/quokka/blob/1b572e70c1ab71d40e8c7f8650d9c90f4fcbe0df/src/EOS.hpp#L133

psharda:
`const amrex::Real /*Tgas*/`  - does it mean Tgas is accessible to the funtion even though there is a  /* */ around it? Also, can't another file call this function without giving it a temperature?

BenWibking:
> `const amrex::Real /*Tgas*/` - does it mean Tgas is accessible to the funtion even though there is a /* */ around it? Also, can't another file call this function without giving it a temperature?

You just have to uncomment it. It's currently commented with `/* */` because it was previously unused. A temperature should be passed to this function anywhere it's currently used -- I don't think that's an issue.

psharda:
> You just have to uncomment it. It's currently commented with `/* */` because it was previously unused. A temperature should be passed to this function anywhere it's currently used -- I don't think that's an issue.

Yes, this is what I wanted to not change. But it's a small change anyway so okay, I will do that and then we don't need this PR.

PR #1301, opened 2023-07-26
Compute pressure for primordial chem EOS

psharda:
Added the calculation of pressure for the primordial chem EOS.

### DIFF
--- a/EOS/primordial_chem/actual_eos.H
+++ b/EOS/primordial_chem/actual_eos.H
@@ -266,6 +266,11 @@ void actual_eos (I input, T& state)
         state.e = eint;
     }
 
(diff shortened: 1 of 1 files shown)


EOS/primordial_chem/actual_eos.H
```diff
@@ -266,6 +266,11 @@ void actual_eos (I input, T& state)
         state.e = eint;
     }
 
+    Real pressure = state.rho * eint / sum_gammasinv;
```
zingale, line 269: you can move the computation of the pressure inside the `if constexpr` to save yourself the compute if it doesn't exist

maxpkatz (approved):

PR #1302, opened 2023-07-27
Compute thermodynamic derivatives and implement dens+pres mode in primordial chem EOS

psharda:
This PR adds the `eos_input_rp` mode for primordial chem EOS. It also computes the sound speed and stores it in `state.cs` if pressure is defined.

### DIFF
--- a/EOS/gamma_law/actual_eos.H
+++ b/EOS/gamma_law/actual_eos.H
@@ -265,7 +265,7 @@ void actual_eos (I input, T& state)
 
             // sound speed
             state.cs = std::sqrt(eos_gamma * state.p * rhoinv);
-            state.d2pdr2_s = 0.0;
+            state.d2pdr2_s = state.cs * state.cs * (std::pow(state.cs, 2) / state.p - rhoinv);
         }
     }
 
--- a/EOS/gamma_law/actual_eos.H
+++ b/EOS/gamma_law/actual_eos.H
@@ -265,6 +265,7 @@ void actual_eos (I input, T& state)
 
             // sound speed
             state.cs = std::sqrt(eos_gamma * state.p * rhoinv);
+            state.d2pdr2_s = 0.0;
         }
     }
 
--- a/EOS/gamma_law/actual_eos.H
+++ b/EOS/gamma_law/actual_eos.H
@@ -265,7 +265,7 @@ void actual_eos (I input, T& state)
 
             // sound speed
             state.cs = std::sqrt(eos_gamma * state.p * rhoinv);
-            state.G = 0.5 * (1.0 + (state.rho / state.p) * state.cs * state.cs);
+            state.G = 0.5 * (1.0 + eos_gamma);
         }
     }
 
--- a/EOS/gamma_law/actual_eos.H
+++ b/EOS/gamma_law/actual_eos.H
@@ -265,7 +265,7 @@ void actual_eos (I input, T& state)
 
             // sound speed
             state.cs = std::sqrt(eos_gamma * state.p * rhoinv);
-            state.G = 0.5 * (1.0 + (rho / state.p) * state.cs * state.cs);
+            state.G = 0.5 * (1.0 + (state.rho / state.p) * state.cs * state.cs);
         }
     }
 
--- a/EOS/gamma_law/actual_eos.H
+++ b/EOS/gamma_law/actual_eos.H
@@ -265,7 +265,7 @@ void actual_eos (I input, T& state)
 
             // sound speed
             state.cs = std::sqrt(eos_gamma * state.p * rhoinv);
-            state.d2pdr2_s = state.cs * state.cs * (std::pow(state.cs, 2) / state.p - rhoinv);
+            state.G = 0.5 * (1.0 + (rho / state.p) * state.cs * state.cs);
         }
     }
 
--- a/EOS/gamma_law/actual_eos.H
+++ b/EOS/gamma_law/actual_eos.H
@@ -198,7 +198,7 @@ void actual_eos (I input, T& state)
 
     // Compute the pressure simply from the ideal gas law, and the
     // specific internal energy using the gamma-law EOS relation.
-    Real pressure = state.rho * state.T * (C::k_B / (state.mu * m_nucleon));
+    Real pressure = state.rho * state.T * C::k_B / (state.mu * m_nucleon);
     Real energy = pressure / (eos_gamma - 1.0) * rhoinv;
     if constexpr (has_pressure<T>::value) {
         state.p = pressure;
--- a/EOS/primordial_chem/actual_eos.H
+++ b/EOS/primordial_chem/actual_eos.H
@@ -278,7 +278,7 @@ void actual_eos (I input, T& state)
         state.dpdT = pressure / temp;
         state.dpdr = pressure / dens;
         state.cs = std::sqrt((1.0 + 1.0/sum_gammasinv) * state.p /state.rho);
-        state.G = 0.5 * (1.0 + (state.rho / state.p) * state.cs * state.cs);
+        state.G = 0.5 * (1.0 + (1.0 + 1.0/sum_gammasinv));
     }
 
     Real dedT = sum_gammasinv * sum_Abarinv * gasconstant;
--- a/EOS/primordial_chem/actual_eos.H
+++ b/EOS/primordial_chem/actual_eos.H
@@ -214,11 +214,16 @@ void actual_eos (I input, T& state)
 
     case eos_input_rp:
         // dens, pressure, and xmass are inputs
-#ifndef AMREX_USE_GPU
-        amrex::Error("eos_input_rp is not supported");
-#endif
 
-          break;
+        if constexpr (has_pressure<T>::value) {
+            dens = state.rho;
+
+            // stop the integration if the pressure < 0
+            AMREX_ASSERT(state.p > 0.);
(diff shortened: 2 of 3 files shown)


EOS/primordial_chem/actual_eos.H
```diff
@@ -214,11 +214,16 @@ void actual_eos (I input, T& state)
...
-#ifndef AMREX_USE_GPU
-        amrex::Error("eos_input_rp is not supported");
-#endif
 
-          break;
+        if constexpr (has_pressure<T>::value) {
+            dens = state.rho;
+
+            // stop the integration if the pressure < 0
+            AMREX_ASSERT(state.p > 0.);
```
BenWibking, line 222: `AMREX_ASSERT` only works when compiled in debug mode. I'm not sure whether it makes sense to use `amrex::Abort` here or not, which should always work.

EOS/primordial_chem/actual_eos.H
```diff
@@ -269,6 +274,8 @@ void actual_eos (I input, T& state)
     if constexpr (has_pressure<T>::value) {
         Real pressure = state.rho * eint / sum_gammasinv;
         state.p = pressure;
+
+        state.cs = std::sqrt((1.0 + 1.0/sum_gammasinv) * state.p /state.rho);
```
BenWibking, line 278: I don't see how this is right. When there is a single species, this should reduce to $\sqrt{\gamma P / \rho}$, right?

psharda: Yes. `sum_gammasinv = 1 / (gamma - 1)`

BenWibking: Ah, I missed the `- 1`. 👍

EOS/primordial_chem/actual_eos.H
psharda: We could. I just used it because thats what we use above in `eos_input_re`

EOS/gamma_law/actual_eos.H
```diff
@@ -198,7 +198,7 @@ void actual_eos (I input, T& state)
 
     // Compute the pressure simply from the ideal gas law, and the
     // specific internal energy using the gamma-law EOS relation.
-    Real pressure = state.rho * state.T * (C::k_B / (state.mu * m_nucleon));
+    Real pressure = state.rho * state.T * C::k_B / (state.mu * m_nucleon);
```
psharda, line 201: This change, although trivial, is necessary to ensure an _exact_ cancellation while calculating pressure from temperature. This is what was causing HydroContact problem in Quokka to fail: https://github.com/quokka-astro/quokka/pull/335

BenWibking: Nice, floating point issues are not easy to find and fix!

EOS/primordial_chem/actual_eos.H
maxpkatz: You can use `AMREX_ALWAYS_ASSERT` if you want this to be used in release builds as well, but in the other equations of state we don't have checks like this. Instead for us the enforcement on positive quantities happens elsewhere in the application code prior to the EOS being called. Mostly the reason for doing it this way is that there is a cost to having an abort in the compilation path for a GPU build.

I think having the assertion on only in debug builds is fine, but that you probably should also do this check in Quokka in a place where it's more convenient to abort if you find it.

BenWibking: That's good to know about the GPU cost penalty. Is that because of additional register usage?
We should probably rethink how we do error handling of this type in Quokka.

`AMREX_ALWAYS_ASSERT` doesn't work if `NDEBUG` is defined, though, which CMake puts in when doing release builds :(

EOS/gamma_law/actual_eos.H
```diff
@@ -265,6 +265,7 @@ void actual_eos (I input, T& state)
 
             // sound speed
             state.cs = std::sqrt(eos_gamma * state.p * rhoinv);
+            state.d2pdr2_s = 0.0;
```
BenWibking, line 268: This is not zero for an ideal gas.

psharda: what is it?

BenWibking: Which term is this supposed to be in these equations:
257545030-2685e971-3233-414c-b442-f0af99f94f69

BenWibking: The red one is zero for an ideal gas. The blue one is not, which you can see by equating 2.23 and 2.24.

psharda: Okay, yes, you are right. The blue term  is not zero. `state.d2pdr2_s` is $`\left(\partial^2 P / \partial \rho^2\right)_S`$. Since the red term is zero, equating these two gives $`\left(\partial^2 P / \partial \rho^2\right)_S = c^2_{s} \left(\frac{c^2_{s}}{P} - \frac{1}{\rho}\right)`$ for an ideal gas.

BenWibking: I got: $(\partial^2 P / \partial \rho^2)_S = (\gamma + 1) (\gamma P / \rho^2) = (\gamma + 1) \ c_s^2 / \rho$

BenWibking: Would it make sense to have the EOS just return the value of $\mathcal{G}$ instead? Then it's just $(1/2)(\gamma + 1)$ for both primordial chem and gamma_law.

psharda: @BenWibking our expressions are the same except that I get a  negative sign for the second term.

BenWibking: Yes. You have to compute gamma in order to compute the pressure, right? I think using gamma directly woudl avoid additional roundoff error.

psharda: > Would it make sense to have the EOS just return the value of G instead? Then it's just (1/2)(γ+1) for both primordial chem and gamma_law.

Yes maybe that is easier,

EOS/primordial_chem/actual_eos.H
```diff
@@ -275,12 +275,20 @@ void actual_eos (I input, T& state)
         Real pressure = state.rho * eint / sum_gammasinv;
         state.p = pressure;
 
+        state.dpdT = pressure / temp;
+        state.dpdr = pressure / dens;
         state.cs = std::sqrt((1.0 + 1.0/sum_gammasinv) * state.p /state.rho);
+        state.G = 0.5 * (1.0 + (state.rho / state.p) * state.cs * state.cs);
```
BenWibking, line 281: I think it would be simpler to write this as $0.5 * (1.0 + (1.0 + 1.0/sum_gammasinv))$.

There's no need to divide by (rho/P), take the square root, square it, and then multiply by it again. It's possible the compiler will optimize this out, but it might not for low optimization settings.

BenWibking (dismissed):

BenWibking (approved):

BenWibking:
@zingale @maxpkatz are you all happy with this? We can hold off on merging until you get a change to review if you'd like.

interfaces/eos_type.H
```diff
@@ -27,6 +27,7 @@ struct eos_t:eos_base_t {
     amrex::Real dsdr{};
     amrex::Real dpde{};
     amrex::Real dpdr_e{};
+    amrex::Real G{}; // fundamental derivative (Thompson 1971), we use equation 2.24 of Menikoff & Plohr 1989
```
maxpkatz, line 30: Is it necessary for this to go in the eos_base_t?

psharda: But this is in eos_t and not eos_base_t. We use eos_t for primordial chem and chem_eos_t for other Quokka problems.

BenWibking: @maxpkatz Is there a better place we can put it?

maxpkatz: Sorry my comment above should have been less ambiguous: my concern was about adding a new quantity to eos_t which only some of the equations of state fill in. Part of the motivation for creating chem_eos_t was so that Quokka could modify that without affecting any of the other equations of state. So in my ideal world, we would only add this G field to chem_eos_t, and then add a `has_G` function analogous to `has_pressure` that wraps around the filling of this field in the EOS, so that the other applications could continue to use the gamma_law EOS without picking up a new quantity. But I may have misunderstood and didn't realize that Quokka was using eos_t in addition to chem_eos_t.

BenWibking: We still use `eos_t` for the `primordial_chem` network, since we need the species fractions to compute the EOS quantities. I suppose we could add the species fraction array to `chem_eos_t`, but my recollection of the rationale for this was that we didn't want to include that when just using the `gamma_law` EOS.

psharda: Yes, @maxpkatz created `chem_eos_t` so that we can use Microphysics EOS for Quokka problems that have no species.

maxpkatz (approved):

zingale:
catching up on this -- we should add G to all of the other EOSes if it is going to be in the `eos_t`

BenWibking:
That would be useful for us, so we can run the test problems in your EOS paper. Is it straightforward to do for the tabulated EOSes?

maxpkatz:
Not at present, we don't have the second derivatives of pressure in Helmholtz. However, this is something we would likely have to add if we were to re-implement the electron-positron portion of Helmholtz from the Timmes EOS, so we could revisit it then.

PR #1304, opened 2023-08-11
Remove legacy screening interface

maxpkatz:
All networks using this legacy approach have now been converted to use the new interface that doesn't rely on the vector of screening factors.


zingale (approved):

zingale:
we should follow this up with a PR that removes the need for `NETWORK_PROPERTIES` and `NSCREEN`

PR #1315, opened 2023-08-21
CMake: do not generate network_properties twice

psharda:
We were generating `network_properties.H` for primordial chem twice. 

Also changed AMReX related variable names to use lowercase 'e'


psharda:
Hmm, why would the docs build be failing?

BenWibking:
It looks like the docs build was broken by a Python upgrade maybe?

BenWibking:
Ok, docs are fixed now.

BenWibking (approved):

BenWibking:
@zingale since this is a pretty small change that shouldn't affect anything other than the CMake build, I'm going to go ahead and merge unless you have objections.

PR #1319, opened 2023-08-28
Update test_rhs/inputs_ecsn to fully exercise tabular rates

yut23:
Also add it to the test_rhs CI workflow.

These inputs cover all the different interpolation/extrapolation cases in the tabular rate evaluation code.

This should cause a failure in the nightly regression tests, and the benchmark will need to be updated on groot.


aisclark91:
I will approve this request, but the update on groot may need to be updated.

aisclark91 (approved):

PR #1322, opened 2023-09-02
add new H-C-Fe-group network

zingale:
this is based on subch_simple but adds a lot of Fe group


networks/He-C-Fe-group/He-C-Fe-group.py
```diff
@@ -0,0 +1,140 @@
...
+for r in rates_to_remove:
+    print("removing: ", r)
+    _r = core_lib.get_rate_by_name(r)
+    core_lib.remove_rate(_r)
+
+# now create a list of iron group nuclei and find both the
+# ReacLib and weak / tabular rates linking these.
+
+iron_peak = ["n", "p", "he4",
+             "mn51",
```
zhichen3, line 65: in some type1a paper I read, there is some discussion over Mn55, saying that it is only stable isotope of Mn, via Co55 -> Fe55 -> Mn55. So I added that when I did my runs, although it didn't generate a lot of them in the end. Not sure whether we add it or not.

zingale: okay, I'll add it tonight

zingale: added

zhichen3 (approved):

PR #1323, opened 2023-09-08
change initial guess from 1e6 to T_in

zhichen3:


zingale (approved):

zingale:
this could change some results to EOS tol level.  But I think this is better than hardcoding in a number

PR #1334, opened 2023-09-12
Move Helmholtz free energy from chabrier1998 into a helper function

yut23:
This reduces code duplication and lets us reuse it in the NSE solver.

Also pull some common subexpressions into variables so we only calculate them once.

### DIFF
--- a/nse_solver/nse_solver.H
+++ b/nse_solver/nse_solver.H
@@ -16,6 +16,7 @@
 #include <eos_composition.H>
 #include <microphysics_sort.H>
 #include <hybrj.H>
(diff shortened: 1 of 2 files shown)


yut23:
I checked against `test_screening` built with `SCREEN_METHOD=chabrier1998` and `burn_cell` built with `NETWORK_DIR=ase USE_NSE_NET=TRUE SCREEN_METHOD=chabrier1998` and run on `inputs_subch_approx`, and the outputs match exactly.

zhichen3 (approved):

PR #1339, opened 2023-09-23
Remove general_null from primordial chem

psharda:
We don't want to include general_null in the directories for primordial chem


psharda:
ready to be reviewed @zingale

zingale (approved):

PR #1342, opened 2023-09-26
add some breaks in the valid step logic

zingale:
no need to keep checking if we already failed


zingale:
test suite passes:
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2023-09-27-004/index.html

maxpkatz (approved):

PR #1356, opened 2023-10-15
Add CI test that fails if a PR is issued to main

maxpkatz:


maxpkatz:
This is a little tricky. There are several ways to inspect the target branch of the PR. For example, the code below is sufficient. The problem is that if this fails (as intended) when the PR targets main, and the target branch is switched to development, re-running the action will not fix the check. The re-run will not update for the target branch changing, since an action re-run appears to be precisely a re-run of the original action, which at the time it was created was on main, so the re-run also thinks we're still on main. This would resolve itself if a new commit were added though. But we should not want to have to require a new commit just to fix this.

```
$ cat check_pr_branch.yml
name: check PR branch

on:
  pull_request:
    branches:
      - main

jobs:
  check-PR-branch:
    runs-on: ubuntu-latest
    steps:
      - name: PRs should not target main. Use development instead.
        run: exit 1
```

maxpkatz:
This appears to be a symptom of the fact that changing the base branch is actually not an event that triggers a workflow to run within the GitHub [pull_request](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) event API. So either we need to manually trigger some action that updates the PR, such as adding a new commit or closing and reopening the PR, or we need some more clever approach that indirectly solves this problem.

zingale (approved):

yut23:
I'm pretty sure this is working now. It turns out target branch changes fall under the `edited` activity type, which I suppose makes sense after the fact since you have to use the "Edit" button to do that.

zingale (approved):

PR #1362, opened 2023-10-20
update networks, benchmarks, and NSE_index due to specie name capitalization

zhichen3:
corresponding update due to pynucastro/pynucastro#652


zingale:
did we expect `test_rhs` to fail?

zhichen3:
test_rhs result changed probably because we updated the neutrino loss procedure for weak rates

> did we expect `test_rhs` to fail?

probably because of pynucastro/pynucastro#655

zingale:
okay, let me regen that net and update the benchmark in a separate PR

zingale (approved):

PR #1371, opened 2023-10-27
move some Fermi integral coefficients to static arrays

zingale:
no need to set these each time we enter the function


zingale:
moved them back into the functions as const Array1D

maxpkatz (approved):

PR #1373, opened 2023-10-27
move sneut5 recombination section into its own function

zingale:


zingale:
test suite passes:
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2023-10-28-004/index.html

maxpkatz (approved):

PR #1374, opened 2023-10-27
move sneut bremsstrahlung into its own function

zingale:


maxpkatz (approved):

zingale:
tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2023-10-28-005/index.html

PR #1380, opened 2023-11-06
Fix swapped sin and cos in sneut

yut23:


zingale (approved):

zingale:
oh, this went to main.  I'll push it to development too

PR #1385, opened 2023-11-13
don't allow logical runtime parameters

zingale:
when we originally did the Fortran -> C++ conversion, we passed all of the Fortran logicals into C++ ints.  But now that we are completely C++, it doesn't make sense to have logical anymore.  There are some places where we implicitly pass an int to a C++ bool.  This will help track those down and eliminate things.


zingale:
the Castro test will fail -- we need to do the same change to the 1 `_parameters` file in Castro that overrides a parameter here.

maxpkatz (approved):

PR #1389, opened 2023-11-17
New neutrino cooling routine

khanakbhargava:
Adding rates using expressions from Kippenhahn & Weigert. Still working on adding derivatives.

### DIFF
--- a/neutrinos/mod_sneut5.H
+++ b/neutrinos/mod_sneut5.H
@@ -1,0 +1,118 @@
+#ifndef SNEUT5_H
+#define SNEUT5_H
+
+#define _USE_MATH_DEFINES
+
+#include <AMReX_REAL.H>
+#include <AMReX_Array.H>
+
+using namespace amrex;
+
+template <int do_derivatives>
+AMREX_GPU_HOST_DEVICE AMREX_INLINE
+void sneut5(const Real temp, const Real dens, const Real abar,
+            const Real zbar, Real& snu, Real& dsnudt, Real& dsnudd,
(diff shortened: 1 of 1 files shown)


neutrinos/mod_sneut5.H
```diff
@@ -0,0 +1,118 @@
...
+rho = dens;
+snu = 0.0e0_rt;
+mu_e = abar/zbar;
+dsnudt  = 0.0e0_rt;
+dsnudd  = 0.0e0_rt;
+dsnuda  = 0.0e0_rt;
+dsnudz  = 0.0e0_rt;
+
+// constants 
+constexpr Real q_e = 4.8032e-10_rt; // Electron charge cm^3/2 g^1/2 s^-1
```
zingale, line 32: you should be able to get all of these from `fundamental_constants.H`

neutrinos/mod_sneut5.H
```diff
@@ -0,0 +1,118 @@
...
+#define _USE_MATH_DEFINES
+
+#include <AMReX_REAL.H>
+#include <AMReX_Array.H>
+
+using namespace amrex;
+
+template <int do_derivatives>
+AMREX_GPU_HOST_DEVICE AMREX_INLINE
+void sneut5(const Real temp, const Real dens, const Real abar,
```
zingale, line 13: you should give this a different name, and then we need to write a driver that wraps the different implmentations and let's us chose which to use, just like we do with screening

khanakbhargava:
Will start a new request to include autodiff.

PR #1390, opened 2023-11-19
[WIP] update B/A in the SDC+NSE

zingale:


zingale:
this may not be needed since I prefer #1392

PR #1393, opened 2023-11-22
Add subch base network

zhichen3:


zingale:
can you add a description here about what the purpose of this network is

zingale:
nevermind, I found it

This network is a further simplification of subch_simple by
￼excluding N14, F18, Ne21, and Na22.
￼These nuclei were added to include n14(a, g)f18(a, p)ne21 rate sequence,
￼which allows an enhancement to the c12(p, g)n13(a, p)o16 rate due to the
￼proton release. However, we find that the effect is not very significant,
￼and they also caused an artificial endpoint at Na22.

zingale (approved):

PR #1395, opened 2023-11-26
work around a CUDA linker issue with the NSE table

zingale:


maxpkatz (approved):

zingale:
according to NVIDIA this is not a bug.

std::max accepts arguments by reference (TIL) and there is an nvcc limitation where you can only use constexpr host variables in device code if it’s by-value and not by-reference

PR #1400, opened 2023-11-28
wrap SDC+NSE neutrino calls with NEUTRINOS

zingale:
this will allow them to be disabled at compile-time just like we do for a network

### DIFF
--- a/integration/nse_update_simplified_sdc.H
+++ b/integration/nse_update_simplified_sdc.H
@@ -75,11 +75,13 @@ void sdc_nse_burn(BurnT& state, const Real dt) {
     Real abar = state.y[SFX+iabar] / rho_old;
     Real zbar = abar * ye_in;
 
+#ifdef NEUTRINOS
     constexpr int do_derivatives = 0;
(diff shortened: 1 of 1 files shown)


integration/nse_update_simplified_sdc.H
```diff
@@ -75,11 +75,13 @@ void sdc_nse_burn(BurnT& state, const Real dt) {
     Real abar = state.y[SFX+iabar] / rho_old;
     Real zbar = abar * ye_in;
 
+#ifdef NEUTRINOS
     constexpr int do_derivatives = 0;
     sneut5<do_derivatives>(T_in, rho_old, abar, zbar,
                            snu, dsnudt, dsnudd, dsnuda, dsnudz);
 
     Real snu_old = snu;
```
zhichen3, line 83: I think you meant to put `snu_old` outside just like the NSE_NET version below.

zingale: fixed

zhichen3 (approved):

PR #1403, opened 2023-11-30
Update C++ linter settings to speed up cppcheck

yut23:
We were previously testing a bunch of invalid preprocessor configurations, which we can now exclude. Installing all the build tools is unnecessary since the linter step runs inside its own Docker container.


yut23:
This brings the runtime down from ~25-30 minutes to ~6-7 minutes.

PR #1407, opened 2023-12-12
add a method to compute the T derivative of an NSE table quantity

zingale:
this works by differentiating the cubic interpolant

### DIFF
--- a/nse_tabular/nse_table.H
+++ b/nse_tabular/nse_table.H
@@ -133,12 +133,13 @@ Real cubic(const Real* xs, const Real* fs, const Real dx, const Real x) {
     // to the data (xs, fs)
     // we take x_i to be x[1]
 
-    Real a = (3 * fs[1] - 3 * fs[2] + fs[3] - fs[0]) / (6 * std::pow(dx, 3));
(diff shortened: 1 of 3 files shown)


nse_tabular/nse_table.H
```diff
@@ -138,6 +142,30 @@ Real cubic(const Real* xs, const Real* fs, const Real dx, const Real x) {
...
+AMREX_GPU_HOST_DEVICE AMREX_INLINE
+Real cubic_deriv(const Real* xs, const Real* fs, const Real dx, const Real x) {
+
+    // fit a cubic of the form
+    // f(x) = a (x - x_i)**3 + b (x - x_i)**2 + c (x - x_i) + d
+    // to the data (xs, fs)
+    // we take x_i to be x[1]
+    // then return dfdx = 3 a (x - x_i)**2 + 2 b (x - x_i) + c
+
+    Real a = (3 * fs[1] - 3 * fs[2] + fs[3] - fs[0]) / (6 * std::pow(dx, 3));
```
maxpkatz, line 159: I think the integer 3 gets cast to a float in std::pow, so this will take a much slower path compared to just writing (dx * dx * dx).

zingale: these have been switched to amrex::Math::powi<N>(x)

yut23 (approved):

PR #1408, opened 2023-12-14
make sure const T evolution

zhichen3:
make sure const temperature evolution when call_eos_in_rhs=false


zingale:
one tests is affected: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2023-12-14-001/test_react_C-aprox19-NSE-VODE.html

zingale (approved):

PR #1421, opened 2023-12-19
add another Castro action that tests against AMReX development

zingale:
the current one tests against the same tag as Microphysics, so if we adapt something here to keep pace with AMReX, then we will fail the current test.

### DIFF
--- a/.github/workflows/castro-development.yml
+++ b/.github/workflows/castro-development.yml
@@ -1,0 +1,89 @@
+name: Castro development
+
+on: [push, pull_request]
+
+concurrency:
+  group: ${{ github.ref }}-${{ github.head_ref }}-apps
+  cancel-in-progress: true
+
+jobs:
+  castro-development:
+    name: Castro development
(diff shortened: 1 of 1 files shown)


.github/workflows/castro-development.yml
```diff
@@ -0,0 +1,89 @@
...
+jobs:
+  castro-development:
+    name: Castro development
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Get Latest Release Tag
+      run: |
+        AMREX_TAG=$(wget https://github.com/AMReX-Codes/amrex/releases/latest 2>&1 | grep Location: | awk '{print $2}' | awk -F/ '{print $NF}')
+        echo "AMREX_TAG=$AMREX_TAG" >> $GITHUB_ENV
```
yut23, line 18: I think this can be removed

zingale: done

yut23 (approved):

Issue #1394, opened 2023-11-22
avoid "extern" device global variables


BenWibking:
As part of debugging https://github.com/AMReX-Astro/Microphysics/issues/1386, it was proposed that Microphysics be re-written to avoid carrying around state in device global variables by moving the state to a struct that could passed to individual functions.

It would be useful to have a list of files that contain device globals. Is anything else needed for this, conceptually?

@maxpkatz @zingale

BenWibking:
Are all of the relevant variables in this list?
https://github.com/search?q=repo%3AAMReX-Astro%2FMicrophysics+extern+AMREX_GPU_MANAGED&type=code&p=1

PR #1422, opened 2023-12-19
add struct support for runtime_parameters.py

zingale:
this will be used to make runtime parameters no longer managed

### DIFF
--- a/util/build_scripts/runtime_parameters.py
+++ b/util/build_scripts/runtime_parameters.py
@@ -99,7 +99,6 @@ class Param:
 
         ostr = ""
 
-        print("here: ", self.cpp_var_name)
         if not self.debug_default is None:
             ostr += "#ifdef AMREX_DEBUG\n"
             ostr += f"{' '*indent}{self.get_cxx_decl()} {self.cpp_var_name}{{{self.default_format(lang='C++', debug=True)}}};\n"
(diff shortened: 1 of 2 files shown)


util/build_scripts/runtime_parameters.py
```diff
@@ -110,27 +127,51 @@ def get_default_string(self):
...
+
+        ostr = ""
+        if self.is_array():
+            # we need to create an amrex::Vector to read and then
+            # copy into our managed array
+            ostr += "\n"
+            ostr += f"        amrex::Vector<{self.get_cxx_decl()}> {self.name}_tmp({self.size}, {self.default_format(lang='C++')});\n"
+            ostr += f"        if (pp.queryarr(\"{self.name}\", {self.name}_tmp, 0, {self.size})) {{\n"
+            ostr += f"            for (int n = 0; n < {self.size}; n++) {{\n"
+            ostr += f"                {cname}{struct_name}.{self.namespace}{self.namespace_suffix}.{self.cpp_var_name}[n] = {self.name}_tmp[n];\n"
```
zhichen3, line 170: I checked issue 2685, the `{self.namespace}{self.namespace_suffix}` here is meant to be a struct like `hydro_params` in the future? I think calling it a namespace might be confusing.

zingale: I agree, but I am trying to reuse the existing functionality, which at the moment was written for the parameters to be global, each in their own namespace.  I suppose I can just create duplicate fields in the class with different names.

But I envision the eventual goal will be to completely remove the support for the global variables, in favor of the structs, and there will be a lot of cleaning at that point.  So I think of this as a first perturbation in that direction.

zhichen3 (approved):

BenWibking:
Partially fixes https://github.com/AMReX-Astro/Microphysics/issues/1394.

maxpkatz (approved):

PR #1423, opened 2023-12-19
fix concurrency for castro-development

zingale:


.github/workflows/castro-development.yml
```diff
@@ -32,7 +32,7 @@ jobs:
       uses: actions/cache@v3
       with:
         path: ~/.cache/ccache
-        key: ccache-${{ github.workflow }}-${{ github.job }}-git-${{ github.sha }}
+        key: ccache-${{ github.workflow }}-${{ github.job }}-git-${{ github.sha }}-development
```
yut23, line 35: Looks like this isn't needed, since `github.workflow` and `github.job` are different between the workflows

zingale: removed

yut23 (approved):

PR #1426, opened 2023-12-21
Replace std::pow with integer powers to amrex::Math::powi

zhichen3:
With updated network to reflect changes in https://github.com/pynucastro/pynucastro/pull/709.

Added a script and CI to detect cases where we can potentially use powi instead of std::pow for integer powers.


zingale:
test suite: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2023-12-22-001/index.html

screening/screen.H
```diff
@@ -715,8 +715,8 @@ void chugunov2009 (const plasma_state_t& state,
     Real Gamma_12 = Gamma_e * z1z2 / scn_fac.ztilde;
 
     // Coulomb barrier penetrability, eq. 10
-    constexpr Real tau_factor = gcem::pow(
-            27.0_rt/2.0_rt * gcem::pow(M_PI*C::q_e*C::q_e/C::hbar, 2) / (C::n_A*C::k_B),
```
zingale, line 719: I think we need this to stay with gcem since this is constexpr and we need it evaluated by the compiler

screening/screen_data.H
```diff
@@ -106,7 +106,7 @@ namespace scrn {
 #elif SCREEN_METHOD == SCREEN_METHOD_chugunov2007
         scn_fac.ztilde = 0.5_rt * (gcem::pow(z1, 1.0_rt / 3.0_rt) +
                                    gcem::pow(z2, 1.0_rt / 3.0_rt));
-        scn_fac.ztilde3 = gcem::pow(scn_fac.ztilde, 3);
```
zingale, line 109: this should stay gcem

zingale: actually, the amrex version is constexpr

screening/screen.H
zingale: it's okay since amrex powi is constexpr

zingale (approved):

zingale:
@psharda @BenWibking do you want to look at the primordial chem changes?
the idea here is that using the templated pow for integer powers should be faster than the library call.

BenWibking:
> @psharda @BenWibking do you want to look at the primordial chem changes? the idea here is that using the templated pow for integer powers should be faster than the library call.

Is this for debug builds?

zingale:
for any build.  It was pointed out to us that C++ no longer has an integer power variant of pow().  It looks like that was removed with C++11:

https://en.cppreference.com/w/cpp/numeric/math/pow

BenWibking:
> for any build. It was pointed out to us that C++ no longer has an integer power variant of pow(). It looks like that was removed with C++11:
> 
> https://en.cppreference.com/w/cpp/numeric/math/pow

On x86, gcc optimizes it out anyway: https://godbolt.org/z/z3nvW1vqM.

Does this happen for GPU builds? I unfortunately can't parse nvptx to figure out if it's doing this in CUDA.

BenWibking:
Ah, well. It looks like gcc only optimizes `pow(x, 1)` and `pow(x, 2)` to avoid calls to `pow`. For higher integer powers, it apparently doesn't optimize it out: https://godbolt.org/z/nd1rqxG8b.

maxpkatz:
In general, NVIDIA will replace std::pow with the device functions `powf` or `pow` which only accept float/double for the exponent argument. This will not be optimized away for the integer case (at least in general; there may be special cases, but I don't remember there being any).

BenWibking:
> @psharda @BenWibking do you want to look at the primordial chem changes? the idea here is that using the templated pow for integer powers should be faster than the library call.

It looks like our unit test passes, so I'm fine with this.

We do have a lot of negative integer powers - just to be sure, does it handle those correctly?

zingale:
yes:

```
 189 //! Return pow(x, Power), where Power is an integer known at compile time                             
 190 template <int Power, typename T,                                                                      
 191     typename = typename std::enable_if<!std::is_integral<T>() || Power>=0>::type>                     
 192 AMREX_FORCE_INLINE                                                                                    
 193 constexpr T powi (T x) noexcept                                                                       
 194 {                                                                                                     
 195     if constexpr (Power < 0) {                                                                        
 196         return T(1)/powi<-Power>(x);                                                                  
 197     } else if constexpr (Power == 0) { ```

Issue #1396, opened 2023-11-26
speed up chabrier1998 screening  by adding `` entries


zingale:
Currently, in `chabrier1998`, we do:

```
  Real Gamma1 = Gamma_e * std::pow(scn_fac.z1, 5.0_rt/3.0_rt);                                                                                          
  Real Gamma2 = Gamma_e * std::pow(scn_fac.z2, 5.0_rt/3.0_rt);                                                                                          
  Real Gamma12 = Gamma_e * std::pow(zcomp, 5.0_rt/3.0_rt);   
```

we should move these pow's into `screen_factors_t`, perhaps only enabled if we are using this screening routine.  This will save on a lot of expensive pows

yut23:
The quantum correction terms also use a lot of pows, which we should probably switch to nested evaluation like in screen5.

zhichen3:
I'm actually debating on whether those terms are needed. These quantum correction terms were not in the original chabrier&potekhin paper. They were introduced in Calder's paper, so I added them. Removing them might give better convergence with NSE states too, although I never tested it.

zingale:
I think that this has been addressed now

PR #1428, opened 2023-12-22
optional quantum correction terms for chabrier1998

zhichen3:
added a runtime parameter to make quantum correction terms in chabrier1998 screening optional. This is because in NSE_solver quantum terms are not included. 

Also fixed some inconsistent indentations

This should also speed up this screening a little bit (issue #1396), although we already switched integer pows to powi.

pr #1426 should be reviewed before this.


zingale:
chug09 fails to compile:
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2023-12-31-003/test_screening_chug09.html

zingale:
do we expect such large diffs from this PR?
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2023-12-31-003/test_screening_chabrier1998.html

zhichen3:
> chug09 fails to compile: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2023-12-31-003/test_screening_chug09.html

oh there is a bug in the script I wrote that it wouldn't pick up the correct integer power if there are multiple commas (such as pow inside a pow). That was the case for gcem::pow in chugunov2009, and you ran the test suite before I added that so it didn't pick it up.

zhichen3:
> do we expect such large diffs from this PR? http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2023-12-31-003/test_screening_chabrier1998.html

Is it possible to run it with `screening.enable_chabrier1998_quantum_corr=1`? Results should be identical if things are correct.

zingale:
yep, I'll do that in the suite

zingale:
tests pass now:
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2024-01-01-002/index.html

zingale (approved):

PR #1432, opened 2024-01-01
[WIP] fix powi script and bug in chugunov2009 screening

zhichen3:
Fixed powi script to test for nested pow, I changed it so that it matches the inner most pow at a time and iterates over to replace all nested pows.

Also fix the bug introduced in chugunov2009 from the previous pr.


zingale (approved):

zhichen3:
was meant to be a [WIP] because the script doesn't catch all nested pow cases I think. I'll follow up with another PR in the future to fix it.

PR #1433, opened 2024-01-02
create a struct to hold the Microphysics runtime parameters

zingale:
This now creates a header extern_type.H with a struct extern_t that holds all
of the runtime parameters for Microphysics.  The struct is filled on initialization
and returned by init_extern_parameters().  This can replace the global version
of the runtime parameters.

This also revives the test_parameters unit test to show how to access things.


psharda:
Looks like we will also need to modify CMakeLists to reflect the changes. @BenWibking ?

unit_test/test_parameters/_parameters
yut23, line 1: This need to be updated since #1536

util/build_scripts/write_probin.py
```diff
@@ -255,26 +250,55 @@ def write_probin(param_files,
...
+
+
 def main():
 
     parser = argparse.ArgumentParser()
     parser.add_argument('-o', type=str, default="", help='out_file')
     parser.add_argument('--pa', type=str, help='parameter files')
     parser.add_argument('--use_namespaces', action="store_true",
-                        help="put parameters in namespaces")
+                        help="[deprecated] put parameters in namespaces")
```
yut23, line 293: It would be good to print some message if `--use_namespaces` is passed

zingale: agreed.  To be honest, I forgot why I deprecated it, but the answer is that because it is the default now.
So I just want to remove it in the future.  I'll add a warning.

unit_test/test_parameters/_parameters
zingale: updated

util/build_scripts/write_probin.py
zingale: deprecation added

zingale:
tests pass:
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2024-05-07-001/index.html

yut23 (approved):

PR #1435, opened 2024-01-02
remove reinterpret_cast in rhs.H

zingale:


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2024-01-02-003/index.html

zingale:
I think that part of the issue here is that `is_jacobian_term_used<>()` checks if a term is Jacobian term is used by comparing to the 0-based species, so when we change the loops to start at 1 instead of 0, we need to shift the indexing used in `is_jacobian_term_used`

zingale:
all tests pass with this change:
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2024-03-09-002/index.html

networks/rhs.H
```diff
@@ -273,19 +273,17 @@ constexpr int is_jacobian_term_used ()
 }
 
 AMREX_GPU_HOST_DEVICE AMREX_INLINE
-void dgesl (RArray2D& a1, RArray1D& b1)
+void dgesl (RArray2D& a, RArray1D& b)
```
yut23, line 276: `a` can be made const here

yut23 (approved):
I went through and did the index shifts for myself, and got the same code.

networks/rhs.H
zingale: updated

PR #1436, opened 2024-01-02
fix a loop bound in the templated rhs.H linear algebra

zingale:
If we compare to `util/linpack.H`, then the inner loop in dgesl where we are 
solving u * x = y should go from: 1 <= j <= k-1, so when we shift to 0-based indexing,
it should be 0 <= j <= k-2, or `constexper_for<0, k-1>`


zingale:
this just seems to make things worse

PR #1444, opened 2024-01-05
Disable SVML AVX-512 routines in the pyna update script

yut23:
This should help with the spurious roundoff diffs when regenerating the networks.


yut23:
Handy tip: you can update a single network by running `../update_pynucastro_nets.py` from inside its directory.

zingale (approved):

PR #1446, opened 2024-01-10
start of a simple 2-T gamma-law EOS

zingale:
This assumes that there are only 2 species: ions and neutrals, and that they have
the same mass.  Electrons are assumed massless.  There is a single gamma for
all species.  ions and neutrals are grouped together as "heavies" and have one
temperature and electrons have a different temperature.  This requires that aux
data come in with the heavy and electron specific energies.

### DIFF
--- a/EOS/gamma_law_2T/Make.package
+++ b/EOS/gamma_law_2T/Make.package
@@ -1,0 +1,1 @@
+CEXE_headers += actual_eos.H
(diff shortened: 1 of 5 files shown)


zingale:
here are my notes on the thermodynamics: 
https://www.overleaf.com/read/qtyhykwkcbtr#bcec75

PR #1453, opened 2024-01-27
replace asctime -> strftime

zingale:
asctime is unsafe

### DIFF
--- a/unit_test/write_job_info.cpp
+++ b/unit_test/write_job_info.cpp
@@ -46,11 +46,11 @@ void write_job_info(const std::string& dir) {
   jobInfoFile << " Plotfile Information\n";
   jobInfoFile << PrettyLine;
 
-  time_t now = time(nullptr);
-
-  // Convert now to tm struct for local timezone
-  tm* localtm = localtime(&now);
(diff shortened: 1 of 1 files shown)


zingale:
this gives the identical output

unit_test/write_job_info.cpp
```diff
@@ -46,11 +46,11 @@ void write_job_info(const std::string& dir) {
   jobInfoFile << " Plotfile Information\n";
   jobInfoFile << PrettyLine;
 
-  time_t now = time(nullptr);
-
-  // Convert now to tm struct for local timezone
-  tm* localtm = localtime(&now);
-  jobInfoFile   << "output date / time: " << asctime(localtm);
+  const std::time_t now = time(nullptr);
+  char buf[64];
```
BenWibking, line 50: Why 64?

BenWibking (approved):

unit_test/write_job_info.cpp
zingale: cause I just copied from cppreference.com
https://en.cppreference.com/w/cpp/chrono/c/asctime

PR #1234, opened 2023-06-19
standardize integrator error codes

zingale:
this will allow us to more easily template the actual_integrator routines

PR #1472, opened 2024-02-07
Fix exit code check in VODE error message

yut23:


zingale:
indeed, I think I missed this in #1234

zingale (approved):

zingale:
can you also fix it in `actual_integrator_simplified_sdc.H`?

zingale (approved):

PR #1476, opened 2024-02-10
remove the templating for the burn retry

zingale:
it greatly increased the compilation time
now we always make a backup of the state befor ethe burn


zingale:
this does seem to cut down on the build time quite a bit and doesn't seem to affect the run time for the tests.

zingale:
I don't think this is needed anymore now that we did #1476

Issue #1250, opened 2023-07-02
for scale_system, the numerical Jacobian scaling is not correct


zingale:
we are calling `actual_rhs()`, so there is no scaling done in the RHS wrapper for us.  But for the energy derivatives,  I think we are only scaling the terms that are d(edot)/dX, and not the terms that are d(Xdot)/de.

Issue #1242, opened 2023-06-29
for Strang scale_system are we doing the dedot/de term correctly


zingale:
When we are doing Strang integration, we have a Jacobian element of the form d(edot)/de.  With scale_system, we are integrating y = (X, e/e_scale), so we need to scale the energy Jacobian elements.  But for this one, we currently are just doing d(edot)/de / e_scale, but this is not dimensionless.  We should scale by (e / e_scale) I think.

The d(edot)/dX terms are okay.

PR #1479, opened 2024-02-12
work on addressing some issues with the Jacobian scaling

zingale:
this applies if we run with scale_system=1
also as part of this change, allow the Circle theorem to use the numerical Jacobian.


zingale:
this addresses #1242 and #1250

zingale:
this seems to fix the scaling.  For example, on `development`, running VODE with `test_react` using `aprox13` and `inputs_aprox13` with:
```
./main3d.gnu.ex inputs_aprox13
```
I get:
```
Run time = 1.375719445
min number of rhs calls: 6
avg number of rhs calls: 116
max number of rhs calls: 3932
min number of steps: 3
avg number of steps: 94
max number of steps: 3767
```

If I turn on scaling:
```
./main3d.gnu.ex inputs_aprox13 integrator.scale_system=1
```
then the integration is really bad:
```
Run time = 19.50141599
min number of rhs calls: 6
avg number of rhs calls: 1611
max number of rhs calls: 109891
min number of steps: 3
avg number of steps: 733
max number of steps: 46777
```

But if I run with scaling on this PR, I get:
```
Run time = 1.435576058
min number of rhs calls: 6
avg number of rhs calls: 117
max number of rhs calls: 3932
min number of steps: 3
avg number of steps: 94
max number of steps: 3767
```
almost the same as the "no scaling" result

maxpkatz (approved):

Issue #1482, opened 2024-02-14
NSE + SDC update is missing some energy from weak interactions


zingale:
We are currently computing the energy release from NSE using the average binding energy / nucleon, <B/A>, as:

eps = N_A d<B/A> / dt

but this misses the energy from the conversion of a neutron to a proton (since that is not contained in the baryon binding energy).  We need to add a term that looks like:

- N_A [ (m_p + m_e) - m_n] c^2 dYe/dt

this is consistent with Seitenzahl et al. 2009, Eq. 20.

zingale:
Here are some notes deriving this

[naw.pdf](https://github.com/AMReX-Astro/Microphysics/files/14284882/naw.pdf)

zingale:
this has been fixed

PR #1483, opened 2024-02-15
add missing weak rate energy to NSE update

zingale:
This accounts for the fact that when we just do the binding energy, we don't capture the energy difference from converting protons to neutrons.  That term looks like N_A (m_n - (m_p + m_e) c**2 dY_e/dt.  This adds that to the SDC
update.


zingale:
closes #1482

zingale:
this seems to work really well

zhichen3 (approved):

PR #1404, opened 2023-12-07
[WIP] start of a new NSE table interface

zingale:

PR #1486, opened 2024-02-19
fix compilation of nse_table_cell

zingale:


zingale:
noticed this when looking over #1484

yut23 (approved):
For future reference, this was changed in #1404

Issue #1499, opened 2024-03-03
indexing issue in `is_jacobian_used()`?


zingale:
in the templated linear algebra, `rhs.H`, we use `reinterpret_cast` to change the array indexing to be 0-based, but the linear algebra routines use `is_jacobian_term_used<>()`, which checks the 1-based species indices defined in `namespace Species`, so it appears that we have a issue with the indexing.

I think that we need to offset the indexing in `is_jacobian_term_used<>()` by 1.  Note: this will only affect Strang, since with SDC we always return 1.

This could explain some of the differences seen in #1435

zingale:
turns out that all of the logic was wrong.  See #1502.

PR #1500, opened 2024-03-03
Fix issues with is_jacobian_term_used()

maxpkatz:
There were two problems here. First, the logic within the function was incorrect: a Jacobian term is used if *both* spec1 and spec2 are connected to the same rate, rather than if *either* species is used. Second, this function is indexing the species using 1-indexing, but we were calling it with zero-indexed species inside the linear algebra.

### DIFF
--- a/networks/rhs.H
+++ b/networks/rhs.H
@@ -263,7 +263,7 @@ constexpr int is_jacobian_term_used ()
             }
         }
 
-        if (is_spec_1_used || is_spec_2_used) {
+        if (is_spec_1_used && is_spec_2_used) {
             term_is_used = 1;
         }
     });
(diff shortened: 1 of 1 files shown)


zingale:
closes #1499

zingale:
these changes cause `test_linear_algebra` to fail:

```
original x and x from the solve (RHS:: solve): 
                   1    7.162462030968882
                  20    13.97277986846644
                   3   -0.822910119156765
                  40    33.89190438198373
                   5    2.117609026004514
                  60    56.02163984189685
                   7    3.458108032248415
                  80    75.97253851812479
                   9    6.080893721649049
                 100    96.56668190864372
                  11    8.490387506470494
                 120    117.0183692563761
                  13    10.03295745366218
                 140    140.7876345006217
```

zingale:
we worked around this for now in #1502

PR #1503, opened 2024-03-05
remove fast reaction cycle check for NSE_NET check.

zhichen3:
Fast reaction cycle is one of the check that Kushnir introduced in his paper. However this check only works with neutron present in the network. 

In the past, I would just skip this check whenever there is no neutron in the network, which is almost always and it worked fine. 

And I've tested with the network that does have neutron without this check, and it gives more or less the same. So I want to get rid of this.

Also changed the indentation to 4 space to be consistent with the rest of the code.


nse_solver/nse_check.H
```diff
@@ -22,889 +22,378 @@
...
-        ||
-        (aion[n] == aion[current_nuc_ind]+3 &&
-            (zion[n] == zion[current_nuc_ind]+1 || zion[n] == zion[current_nuc_ind]+2))
-        ||
-        (aion[n] == aion[current_nuc_ind]+4 && zion[n] == zion[current_nuc_ind]+2)
-        ) {
+    // if there is neutron in the network
 
-      ++max_nucs;
+    if ((std::abs((r - r_nse) / r_nse) < 0.5_rt)
```
zingale, line 62: you can get rid of an expensive divide here by moving the `r_nse` to the RHS

nse_solver/nse_check.H
```diff
@@ -22,889 +22,378 @@
...
-           || zion[n] == zion[current_nuc_ind]+2))
-         ||
-         (aion[n] == aion[current_nuc_ind]-2 &&
-          (zion[n] == zion[current_nuc_ind]-2 || zion[n] == zion[current_nuc_ind]-1
-           || zion[n] == zion[current_nuc_ind]))
-        ) {
+    // if there is no neutron in the network
 
-      ++max_nucs;
+    if ((std::abs((r - r_nse) / r_nse) < 0.25_rt)
```
zingale, line 70: likewise, you can get rid of the divide

zingale (approved):

PR #1505, opened 2024-03-06
[WIP] GIFT gaussian elimination testing

yut23:
I ported the Fortran code to Python, then added some C++ and Python codegen for us to play with.

It's currently set up to generate code for aprox13. With the current array ordering (species then enuc), it produces the same code as a non-sparse matrix (1118 statements), but if we move enuc before the species, it's a lot better at 302 statements.


BenWibking:
For future reference, the original method paper was: https://dl.acm.org/doi/pdf/10.1145/321556.321565
They claim that in typical cases, the complexity goes from O(N^3) to O(N).

PR #1510, opened 2024-03-13
add Ben and Piyush to .zenodo.json

zingale:


.zenodo.json
```diff
@@ -43,11 +43,21 @@
             "name": "Malone, Chris",
             "orcid": "0000-0002-4045-7932"
         },
+        {
+            "affiliation": "Leiden Observator",
```
BenWibking, line 47: I think there's a "y" missing at the end

zingale: I'm a theorist.  what do you expect?

psharda (approved):

PR #1512, opened 2024-03-15
eliminate some divides in the helmholtz EOS

zingale:


zingale:
tests:
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2024-03-16-001/index.html

I suspect those failures are because I was not fully merged with development.  I'll retest those

zingale:
indeed -- they pass:
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2024-03-16-002/index.html

so all tests pass

maxpkatz (approved):

PR #1515, opened 2024-03-17
extract some common integrator infrastructure

zingale:


zingale:
this looks a bit clumsy, so comments welcomed.  With this, I can make `BackwardEuler` and `RKC` reuse the same `integrator_setup`

zingale:
update: `std::function()` doesn't work on the device, but a simple function pointer does.

zingale:
it appears that function points are not supported by SYCL, so we'll need to use functors (e.g., a struct wrapper with an overload of ())

zingale:
requires #1541

zingale:
this has been refactored so it will be SYCL compliant now

maxpkatz (approved):

PR #1516, opened 2024-03-18
some network_properties.H cleaning

zingale:
put aion, zion, ... into an anonymous namespace instead of making them static 
also disable linting on the switch for lacking a default -- this is hard with the constexpr
these were suggested by clang-tidy
also remove an old Fortran template


networks/general_null/network_header.template
```diff
@@ -96,8 +100,9 @@ namespace NetworkProperties {
 
        amrex::Real z = -1.0;
 
-       switch (spec) {
+       switch (spec) {  // NOLINT(bugprone-switch-missing-default-case)
            @@ZION_CONSTEXPR@@
+
        }
```
yut23, line 106:
```suggestion
       // NOLINTBEGIN(bugprone-branch-clone)
       switch (spec) {  // NOLINT(bugprone-switch-missing-default-case)
           @@ZION_CONSTEXPR@@

       }
       // NOLINTEND(bugprone-branch-clone)
```

yut23 (approved):

yut23:
The clang-tidy docs actually say that anonymous namespaces in header files are an antipattern: https://clang.llvm.org/extra/clang-tidy/checks/misc/use-anonymous-namespace.html https://wiki.sei.cmu.edu/confluence/display/cplusplus/DCL59-CPP.+Do+not+define+an+unnamed+namespace+in+a+header+file
It probably didn't notice because it doesn't recognize .H as a header extension by default.

PR #1532, opened 2024-04-09
have clang-tidy warn about recursion

zingale:
CUDA doesn't seem to like it


yut23 (approved):
`factorial()` from rhs.H comes up in the report, but it's only ever called at compile-time.

PR #1533, opened 2024-04-11
fix some header issues identified by clang-tidy

zingale:
most adding inline to avoid ODR issues


yut23:
I found one more: `nse_solver/make_table/burn_cell.H`

zingale:
fixed

yut23 (approved):

PR #1540, opened 2024-05-07
remove sdc_burn_tol_factor

zingale:
we never use this and it doesn't seem like it will be needed since we've made so many improvements to the integrators.


zingale:
looks like we did this via other commits / PRs

PR #1546, opened 2024-05-09
unify the SDC integrator code paths

zingale:
This removes a lot of the redundancy in the various actual_integrator_sdc.H files


integration/integrator_setup_sdc.H
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    // Fill in the initial integration state.
+
+    burn_to_int(state, int_state);
+
+    // Set the tolerances.
+
+    // we use 1-based indexing inside of the integrator, so we need to
+    // shift the indices SRHO, SFS, etc by 1
```
yut23, line 59: I don't think this comment applies anymore

zingale: indeed, you are right.  I think that's from back when the tolerances were arrays

yut23: The manual indexing was removed in https://github.com/AMReX-Astro/Microphysics/pull/1039

zingale: fixed

yut23 (approved):

Issue #1545, opened 2024-05-08
add second set of NSE criteria


zingale:
This is for tabular NSE

For really high T (like T > 1.e10 K), the composition goes into p + n, so our current check on iron group + He would not trigger the NSE table.  For this reason, we should have a second set of heuristic conditions that also put us into NSE.

zingale:
The current version of `burn_cell` with `inputs_aprox19.nse` has this issue -- we enter NSE in the first substep, but then not after that.

PR #1547, opened 2024-05-10
add a second T check to tablular NSE

zingale:
if we are larger than T_always_nse, then we ignore composition (density still matters)


zingale:
closes #1545

zingale:
note: this changes the output of the burn_cell test where we enter NSE because previously, once we got to T ~ 1.e10, we had no more Fe / He and the heuristics decided we were no longer in NSE, and instead the burner integrated.

zhichen3 (approved):

PR #1548, opened 2024-05-10
add SDC+NSE unit test example

zingale:


khanakbhargava (approved):

.github/workflows/nse_table.yml
yut23, line 1: The first "Print backtrace" step should be removed (so we just have a single copy after all the runs).

yut23 (changes requested):

.github/workflows/nse_table.yml
yut23: Oh, nevermind. I missed that the first is `burn_cell` and the second is `burn_cell_sdc`.

yut23 (approved):

Issue #1542, opened 2024-05-07
remove NSE support from Strang


zingale:
Strang integration doesn't really allow us to evolve the NSE state to second-order accuracy.  We should simply remove NSE as an option from Strang.

PR #1549, opened 2024-05-10
Remove strang nse

zingale:
Strang does not really allow us to do 2nd order easily because we can't capture the evolution of the state during the burn


zingale:
addresses #1542

zingale:
I think that this is good now

integration/Make.package
```diff
@@ -26,11 +26,12 @@ else
   CEXE_headers += integrator_setup_strang.H
 endif
 
+
 ifeq ($(USE_NSE_TABLE), TRUE)
```
zhichen3, line 30: should we include USE_NSE_NET here?

zingale: fixed

zhichen3 (approved):

PR #1559, opened 2024-05-23
some reorganization of the docs into sections

zingale:


zingale:
here's a rendering:

image

zhichen3 (approved):

PR #1561, opened 2024-05-23
add docs on Jacobian caching

zingale:


zingale:
blocked by #1559

sphinx_docs/source/ode_integrators.rst
```diff
@@ -46,6 +46,14 @@ the allowed options are:
 * ``VODE``: the VODE :cite:`vode` integration package.  We ported this
   integrator to C++ and removed the non-stiff integration code paths.
 
+  .. note::
+
+     The VODE integrator uses Jacobian caching when run on a CPU by default.  This
+     can be disabled at runtime by setting ``integrator.use_jacobian_caching = 0``.
+
+     On GPUs, we disable Jacobian caching due to the increased memory needs.  This
+     is disabled at compile-time via the build parameter ``USE_JACOBIAN_CACHING``.
```
zhichen3, line 55: Its currently only disabled by default for `CUDA` I think? Should we change that?

Also I think you mean "disabled by default but can be **enabled** at compile-time via `USE_JACOBIAN_CACHING=TRUE`  "

zingale:
that's a good point -- maybe that explains why HIP performance is not as good as CUDA?

Issue #1555, opened 2024-05-21
change default retry tolerances to be same as normal


zingale:
We should make the default values of the retry tolerances (`retry_rtol_spec`, ...) negative to indicate that they should just be set to be the same as the ones used in the first attempt.  This will make the retry more predictable in case someone forgets to set these.

zingale:
this should wait until after #1515 and #1546 are merged

zingale:
this has been done

PR #1573, opened 2024-05-30
change how the retry tolerances work

zingale:
they now default to -1, in which case the original (non-retry) tolerance will be used.  This makes it easier for a user to use the retry mechanism without having to remember to explicitly set all of the tolerances


zingale:
closes #1555

khanakbhargava (approved):

Issue #1574, opened 2024-05-30
consolidate subch networks


zingale:
Right now we have:

* `subch_approx`
* `subch_base`
* `subch_full`
* `subch_simple`
* `He-C-Fe-group`

plus 2 symlinks.

Do we still need all of these networks?  For instance, I don't think we use `subch_full`.

We should prune these down and make any improvements we thing should be done to the remaining networks.

zingale:
My thinking is that we can remove `subch_approx` and `subch_full` (and the symlinks), and then do something to `subch_simple` to fix the hanging endpoint at Na22.

PR #1578, opened 2024-06-04
Delete subch_full and subch_approx networks.

zhichen3:
delete subch_full and subch_approx to address #1574


sphinx_docs/source/networks.rst
```diff
@@ -241,29 +242,30 @@ forward rate to just be the first rate.  We do not include reverse
 rates for these processes.
 
 
-subch_full
-----------
+subch_full (Deprecated)
```
zingale, line 245: it's not deprecated -- it has been removed

so I think you can just replace these sections with a note saying that the older subch_full and subch_approx networks have been removed in commit ###### (give the hash)

zhichen3: done

zingale (approved):

PR #1583, opened 2024-06-18
implement a fast atan approximation function

zhichen3:
implement a fast atan function using an approximate function for x in [-1, 1], for |x| > 1, the identity `arctan(x) = sign(pi/2, x) - arctan(1/x)` (Thanks to Eric) is used.

Two versions of the approximate functions are included.
1) Efficient Approximations for the Arctangent Function by Rajan 2006: Equation 9: Max error ~0.0015 rad
2) https://stackoverflow.com/questions/42537957/fast-accurate-atan-arctan-approximation-algorithm: Max error ~0.00063 rad

Generally I find the form from stackoverflow is more accurate and also faster. 1) is also discussed in 2).

In general, from testing, atanf is roughly more than 2 times faster than std::atan.


util/approx_math/fast_atan.H
```diff
@@ -0,0 +1,114 @@
+#ifndef MICROPHYSICS_APPROX_MATH_H
```
zingale, line 1: I would suggest calling the header approx_math.H
this way we can add other approximations in the future to the same header

util/approx_math/fast_atan.H
```diff
@@ -0,0 +1,114 @@
...
+    ///
+    /// This calculates atan within [-1, 1] range.
+    ///
+
+    // formula 1
+    // constexpr amrex::Real PI_4 = GCEM_PI / 4.0_rt;
+    // return PI_4 * x;
+
+    // formula 2
+    // constexpr amrex::Real A = 0.0776509570923569_rt;
```
zingale, line 23: can you add the reference to where these expressions came from?

zhichen3:
Not sure why hip failed to compile. Also I kept the volatile qualifier for testing because runtime wouldn't increase with iteration otherwise.

zingale:
I suspect the HIP failure is due to changes in ROCm and there is a namespace clash now.

zingale:
can you try not including `<AMReX_GpuUtility.H>` and instead directly include just what you need?

zingale:
or alternately, include `<AMReX_Gpu.H>`, since that seems to do some namespace magic.

yut23:
It's probably this line: https://github.com/AMReX-Astro/Microphysics/blob/30469dd4b2c8374b5a19708f95f6921748211366/networks/iso7/actual_network.H#L10

yut23:
`atanf()` may conflict with the definition from the standard library. I think `fast_atan()` works fine.

zingale:
indeed, we should get rid of those `using namespace amrex`
I'll try that in a separate PR

zhichen3:
I figured checking nan is unnecessary, so I removed it and also removed `AMReX_GpuUtility.H` since I don't need isnan anymore. I guess this also fixed hip compilation error.

yut23 (approved):

zhichen3:
sounds good to me

PR #1584, opened 2024-06-20
remove using namespace amrex from aprox nets

zingale:


networks/aprox19/nse_table_size.H
yut23, line 1: This still has `using namespace amrex;`

zingale: fixed

yut23 (approved):

PR #1591, opened 2024-06-24
Fast log and pow implementation

zhichen3:
Fast log implementation by inverting fast_exp, roughly 10 times faster than std::log, but it doesn't check bounds like inf and 0. With additional if-statements, it would be slower, but still ~5 times faster than std::log

 This is discussed in:
https://martin.ankerl.com/2007/10/04/optimized-pow-approximation-for-java-and-c-c/

Also adds fast_pow implementation by combining fast_exp and fast_log, this is the vfast_pow function. It is roughly 30 times faster than std::pow but poor accuracy when exponent is >> 1.

If we combine the method of exponentiation by squaring for solving the integer part and only use approximation for the fraction part of the exponent, then we get overall ~3% relative error for all exponents, but its a lot slower than vfast_pow, but still ~5 times faster than std::pow.


zingale (approved):

util/approx_math/approx_math.H
```diff
@@ -270,4 +270,204 @@ amrex::Real fast_exp(const amrex::Real x) {
...
+AMREX_GPU_HOST_DEVICE AMREX_INLINE
+float fast_log(const float x) {
+    // For single precision input
+
+    constexpr float ia = 0.6931471805599453F / gcem::pow(2.0F, 23);
+    constexpr float b = gcem::pow(2.0F, 23) * (255.0F - 0.04367744890362246F);
+
+    // Make sure input is positive
+
+    assert(x >= 0.0F);
```
yut23, line 297: Should this be AMREX_ASSERT?

zingale: yeah, that's probably better

zhichen3: done

PR #1594, opened 2024-06-26
update NSE table check docs

brady-ryan:


zingale:
can you also update the iron-group description in the docs to reflect the changes that you made to consider Z and not the names directly?

zingale (approved):

PR #1595, opened 2024-06-26
(optionally) add T/e correction term to Strang Jacobians

zingale:
this is the dT/dX |_e term that appears in all species
it is now controlled by integrator.correct_jacobian_for_const_e


zingale:
closes #729

zingale:
some Strang data using `burn_cell`.  Build as:

```
make NETWORK_DIR=subch_simple
```

then run the old way:
```
./main3d.gnu.ex inputs_subch_simple integrator.jacobian=1 unit_test.nsteps=1 integrator.correct_jacobian_for_const_e=0
```
and it take 8084 steps.

Running with the new correction, `integrator.correct_jacobian_for_const_e=1` and it takes

6788 steps

zingale:
curiously, with `aprox21`, it goes the other way, and we are slower

zingale:
for He-C-Fe-group, using the correction terms speeds things up by 4x

```
./main3d.gnu.ex inputs_subch_simple integrator.jacobian=1 unit_test.nsteps=1 integrator.correct_jacobian_for_const_e=1 unit_test.tmax=1.e-2
```

104499 without the corrections
24735 with the corrections

zingale:
Here's some `test_react` runs:

* `aprox13`

  with new terms:

  ```
  Run time = 2.016436636
  min number of rhs calls: 6
  avg number of rhs calls: 96
  max number of rhs calls: 1171
  min number of steps: 3
  avg number of steps: 75
  max number of steps: 878
  ```

  without new terms:

  ```
  Run time = 2.397303023
  min number of rhs calls: 6
  avg number of rhs calls: 116
  max number of rhs calls: 3932
  min number of steps: 3
  avg number of steps: 94
  max number of steps: 3767
  ```

* `subch_simple` (using `inputs_aprox13`):

  with new terms:

  ```
  Run time = 101.0353978
  min number of rhs calls: 6
  avg number of rhs calls: 3962
  max number of rhs calls: 147079
  min number of steps: 3
  avg number of steps: 2219
  max number of steps: 79369
  ```

  without new terms:

  ```
  Run time = 100.0271792
  min number of rhs calls: 6
  avg number of rhs calls: 3975
  max number of rhs calls: 157375
  min number of steps: 3
  avg number of steps: 2284
  max number of steps: 86380
  ```

zingale:
here's a Castro run with these changes:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2024-07-12-002/index.html

zingale:
@BenWibking can you make sure that this works for you?

BenWibking:
> @BenWibking can you make sure that this works for you?

It works.

zingale:
can you review it?

BenWibking:
reviewing now

integration/integrator_rhs_strang.H
```diff
@@ -144,6 +144,47 @@ void jac ([[maybe_unused]] const amrex::Real time, BurnT& state, T& int_state, M
...
+    //              | d(rho X2dot)/dX1  d(rho X2dot)/dX2 ... 1/cv d(rho X2dot)/dT |
+    //  SFS-1+nspec |   ...                                                       |
+    //  SEINT       \ d(rho Edot)/dX1   d(rho Edot)/dX2  ... 1/cv d(rho Edot)/dT  /
+    //
+    //                   SFS                                         SEINT
+
+    // now correct the species derivatives
+    // this constructs dy/dX_k |_e = dy/dX_k |_T - e_{X_k} |_T dy/dT / c_v
+
+    if (integrator_rp::correct_jacobian_for_const_e) {
```
BenWibking, line 161: This may need to be disabled for number densities, since it assumes the species variables are mass fractions.

BenWibking (commented):
I think this is only mathematically valid for networks that use mass densities.

BenWibking:
It doesn't crash our network, and it still seems to get the right answer, so the correction must be small or zero in our network.

integration/integrator_rhs_strang.H
zingale: then I'm confused -- that means that this PR will do nothing for the number density case.  Is that your intent?

BenWibking: I think it needs to be rederived for number densities. I can probably do that later this week if it can wait, or it can go in a separate PR.

zingale: let's do that in a separate PR.  I'll wrap this in an if-test for now

zingale:
@BenWibking where did we leave off on this?  Should we merge this and then address the number density case?

BenWibking:
I think we are going to address this in our Python code generator, so that's fine with me.

PR #1603, opened 2024-07-05
Avoid dividing by zero in constexpr linear algebra routines

yut23:
`RHS::dgefa()` now stops if the matrix is singular, and returns an integer that works the same as the `info` parameter for the routine in `linpack.H`. If the LU decomposition fails, it is the index of the row that failed, and is zero if the decomposition succeeded.

This fixes the crash in `Detonation-collision-retry-SDC` in the Castro test suite. The benchmark does need to be updated, since it now stops at step 10 instead of 11, and the plotfile for that step was last updated before we changed the setup in https://github.com/AMReX-Astro/Castro/pull/2806.


zingale:
we'll have to do a separate Castro PR to add the info to the true SDC stuff

zingale (approved):

PR #1611, opened 2024-07-14
include higher taylor terms for fast_atan

zhichen3:
Pointed out by Eric: in screening, atan is used in cases like `gamma - atan(gamma)`, so we should use a higher order term rather than simply letting atan(gamma) ~ gamma when gamma is small.


util/approx_math/approx_math.H
```diff
@@ -39,6 +40,18 @@ amrex::Real fast_atan_1(const amrex::Real x) {
...
+    ///
+    /// If x < 0.378, then using arctan(x) ~ x - x^3/3 + x^5/5
+    /// gives better answer than the approximation below.
+    /// And accuracy increase as x << 0.378.
+    ///
+
+    if (std::abs(x) < 0.378_rt) {
+        // return -0.3333333333_rt*amrex::Math::powi<3>(x) + x;
+        amrex::Real x2 = x*x;
+        return ((0.2_rt*x2 - 0.33333333333_rt)*x2 + 1.0_rt)*x;
```
zingale, line 52: I think you can use more precision for 1/3 -- why not just do
```
constexpr Real third = 1.0_rt/3.0_rt;
```

zhichen3: Yup, fixed.

zingale (approved):

PR #1612, opened 2024-07-15
Update `autodiff::dual` to give the same results as `amrex::Real`

yut23:
In my testing with converting the `aprox*` rates to use autodiff, I found a lot of instances where code templated with `autodiff::dual` would give slightly different values than the same code templated with `amrex::Real`. It turns out to be due to some optimizations autodiff makes to speed up evaluation, which this PR disables to avoid any weird inconsistencies in the future.


zingale:
this is upstreamed here: https://github.com/autodiff/autodiff/pull/352

zingale:
test suite:
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2024-07-16-001/index.html

zingale:
all of the test_rhs diffs are roundoff error level

zingale (approved):

PR #1614, opened 2024-07-17
Use autodiff in templated network rate evaluation

yut23:


yut23:
Timings for `test_aprox_rates`, with a loop added around the `rate_*` calls. `RHS::rhs()` calls the version without derivatives, while the analytic Jacobian calls the version with derivatives. The previous code always calculated the derivatives, hence the same timings on the left.
```
CPU (n_cell=64, loops=10):
- without derivatives:  5.34s ->  4.34s (-1.00s, -19%)
- with derivatives:     5.34s ->  5.71s (+0.37s,  +7%)

CUDA (n_cell=64, loops=500):
- without derivatives:  5.25s ->  4.42s (-0.83s, -16%)
- with derivatives:     5.25s ->  5.30s (+0.05s,  +1%)
```

zingale:
test suite: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2024-07-17-006/index.html

zingale:
do you expect the diffs shown in the test suite?

yut23:
No, I don't think the diffs should be that large. I'm even more confused by the diffs with the numerical Jacobian, as that shouldn't be using autodiff at all.

zingale:
let me rerun the test

yut23:
I found a few optimizations I made that were causing the differences with the numerical Jacobian, and I'm now pretty sure the other big diffs are just compounded from the roundoff diffs in the analytical Jacobian.

zingale:
most test_rhs diffs are now roundoff, except for powerlaw, which is Inf
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2024-07-23-001/index.html

yut23:
This should be ready to merge. The Inf diffs for powerlaw are because we weren't calculating the temperature derivatives at all before.

zingale (approved):

Issue #945, opened 2022-07-11
we zero the Jacobian too many times


zingale:
In `vode_dvjac.H`, we zero the Jacobian before calling the analytic Jacobian, but in the pynucastro networks, we also zero the Jacobian right when we enter the RHS routine.  Note that in the templated RHS networks (in `rhs.H`), we don't zero the Jacobian. 

So we should update the pynucastro net wrappers perhaps to not zero.

zingale:
actually, we do zero the Jacobian in `rhs.H` for the templated nets.  So it looks like in most instances, we are zeroing the Jacobian twice in a row.

`numerical_jacobian.H` does not zero the Jacobian, but it explicitly loops over every element, so this is not an issue.

Note that currently BackwardEuler does not zero the Jacobian before calling the jacobian routine.

zingale:
so we should decide on a convention -- should the integrator or the network handle zeroing out the Jacobian for the analytic Jacobian?

PR #1636, opened 2024-07-30
only zero the Jacobian once

zingale:
now it is the responsibility of the network to ensure that the Jacobian is zeroed, if needed.
We don't have to worry about the numerical Jacobians, since those are computed in a dense
fashion.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,54 @@
+# 24.08
+
+  * autodiff is now used with the templated reaction networks (#1614)
+    + some autodiff clean-ups and derivative fixes (#1604, #1612,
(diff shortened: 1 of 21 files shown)


zingale:
closes #945

zingale:
I am now thinking that perhaps the other approach is better. -- the integrators don't zero at all, and we always assume that the network implementation will handle it.  That makes things more explicit.

zingale:
tests pass: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2024-07-30-002/index.html

yut23 (approved):

Issue #1634, opened 2024-07-29
lower the default ode_max_steps


zingale:
Right now, `integrator.ode_max_steps = 150000` by default.  This is very high, and if we need this many steps, we most likely have a problem (like a bad Jacobian).  In many problems, lowering this to `10000` or so and then turning on `integrator.use_burn_retry` speeds up the execution a lot.

We should lower this default to something smaller like `25000`.  This will force retries (either at the burner or application code level) and should perform better on GPUs where the divergence in the number of steps different zones needs can be a big slowdown.

PR #1638, opened 2024-07-31
lower ode_max_steps and make burn retry on by default

zingale:
experience with science problems has shown that, especially on GPUs, one zone can take ode_max_steps and then fail, forcing the application code to deal with it (e.g. Castro retries).  But often using the burn retry functionality (in particular swapping the Jacobian) makes the zone pass.

This change lowers ode_max_steps from 150000 to 25000 and turns on the retry (which only enables Jacobian swapping by default). This should help with a lot of applications, although lowering ode_max_steps more can help further.

### DIFF
--- a/integration/_parameters
+++ b/integration/_parameters
@@ -44,7 +44,7 @@ MAX_TEMP                     real            1.0e11
 react_boost                 real         -1.e0
 
 # maximum number of timesteps for the integrator
-ode_max_steps              int      150000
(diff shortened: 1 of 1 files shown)


zingale:
closes #1634

zingale:
Castro test suite seems to run fine with this change:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2024-07-31-001/index.html

Issue #1337, opened 2023-09-18
MathArray uses Fortran ordering -- are we looping correctly?


zingale:
In `ArrayUtilities.H`, we define `MathArray2D` with a Fortran-ordering:

```
  86         AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
  87         const Real& operator() (int i, int j) const noexcept {
  88             AMREX_ASSERT(i >= XLO && i <= XHI && j >= YLO && j <= YHI);
  89             return arr[i+j*(XHI-XLO+1)-(YLO*(XHI-XLO+1)+XLO)];
  90         }
```

but when we loop over it (for instance as the Jacobian), we do:

```
 123     // The Jacobian from the nets is in terms of dYdot/dY, but we want
 124     // it was dXdot/dX, so convert here.
 125     for (int n = 1; n <= NumSpec; n++) {
 126         for (int m = 1; m <= neqs; m++) {
 127             pd(n,m) = pd(n,m) * aion[n-1];
 128         }
 129     }
 ```

shouldn't these loops be reversed?

yut23:
It looks like they should be swapped to me.

PR #1639, opened 2024-07-31
reorder some loops over Jacobians

zingale:
the MathArray2D is Fortran-ordered.  Some of the loops should be more efficient if their order is swapped.


zingale:
closes #1337

yut23 (approved):

PR #1642, opened 2024-08-22
Implement ISM chemistry including metals and dust

psharda:
This PR will  build on the existing primordial ISM chemistry network to include a minimal network with metals and dust. Also added are the tests that go with this PR, where the network is integrated at 7 different metallicities and the test solutions are compared against reference solutions.


EOS/primordial_chem/_parameters
```diff
@@ -3,58 +3,58 @@
 eos_gamma_default       real          1.4
 
 # define the specie names, and their masses and gammas
-species_a_name          string     "elec"
-species_a_gamma         real          5./3.
-species_a_mass          real          9.10938188e-28
+species_1_name          string     "elec"
```
psharda, line 6: @zingale why does this change break the primordial chem test? I simply renamed the variable name from 'species_a_name' to 'species_1_name', and so on...

zingale: it seems that `make` returned a non-zero error code.  Maybe this is something you do in cmake?  It compiled fine, but `make` was not happy.

psharda:
@BenWibking do you understand why the metal_chem test failed?

psharda:
> @BenWibking do you understand why the metal_chem test failed? The overflow errors have been corrected now.

Ah nvm, I didn't upload the right file

psharda:
Everything else is good now, except for the CMake test. The test cannot find the Semenov opacity table (stored as a.dat file in networks/metal_chem). Any ideas @BenWibking ?

BenWibking (changes requested):
Well, this is quite a lengthy PR. I did not review the boilerplate code.

There are a few things that can be improved in the interpolation and root-finding done for the dust temperature. Also, I think it's probably not necessary to keep commented-out code in the PR. It makes it even longer and harder to read. If it's there for debugging, you could put it in an `ifdef` instead.

psharda:
> Networks can be unavoidably very big, but IMHO we should find a way to avoid having 50,000+ line PRs to add new chemistry networks in the future.

How can you avoid it? You need everything in place to test the code out. Perhaps one alternative is to create _sub-PRs_ by dividing this PR into several chunks and merging them one by one. But there is no way of testing these sub-PRs, and no guarantee that all the chunks when merged would actually work. ISM chemistry in the presence of metals and dust is inherently complex, more so for our science goals that require going to very low metallicities. 

Note, however, that this is the _simplest_ network that works. FWIF future PRs that add cosmic rays and photochemistry will be smaller, since the base network is in place now.

psharda:
The YML tests also need to be done at all metallicities, not just one. As we have seen, the test might pass at one metallicity but not others.

unit_test/burn_cell_metal_chem/burn_cell.H
```diff
@@ -0,0 +1,378 @@
...
+    //  inmfracs[nn] = spmasses[nn] * state.xn[nn] / state.rho;
+    //  insum += inmfracs[nn];
+    //}
+
+    //for (int nn = 0; nn < NumSpec; ++nn) {
+    //  inmfracs[nn] /= insum;
+    //  // update the number densities with conserved mass fractions
+    //  state.xn[nn] = inmfracs[nn] * state.rho / spmasses[nn];
+    //}
+
```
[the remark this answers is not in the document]
psharda: I wanted to keep the commented out code in question here until I have tried a simulation with quokka, as some of this might come in handy. Most of the code that I used to debug has been removed.

networks/metal_chem/actual_rhs.H
```diff
@@ -0,0 +1,4264 @@
...
+    std::cout << "network Tdust: " << state.aux[0] << std::endl;
+    std::cout << "network ydots: ";
+    for (int i = 1; i < neqs; ++i) {
+        std::cout << "ydot" << i << ": " << ydot(i) << " ";
+    }
+    std::cout << std::endl;
+    std::cout << "network edot: " << edot << std::endl;
+    std::cout << " " << std::endl;
+    */
+
```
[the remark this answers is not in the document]
psharda: Done

networks/metal_chem/actual_rhs.H
```diff
@@ -0,0 +1,4264 @@
...
+    Real A = rhogas * kappaP * user_dust2gas_ratio * aR * C::c_light;
+    Real B = std::pow(Tgas, 0.5) * amrex::Math::powi<2>(nH) * alpha_gd * user_dust2gas_ratio;
+    Real phys_Tcmb = 2.73 * (1.0 + z);
+    Real C = -1.0 * (intJRad + A * amrex::Math::powi<4>(phys_Tcmb) + std::pow(Tgas, 1.5) * amrex::Math::powi<2>(nH) * alpha_gd * user_dust2gas_ratio);
+
+    int iterr = 0;
+    Real Tdold = krome_Semenov_Tdust;
+    Real Tdnew = 0.0;
+
+    while (true) {
```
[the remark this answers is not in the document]
psharda: Done

BenWibking:
> > Networks can be unavoidably very big, but IMHO we should find a way to avoid having 50,000+ line PRs to add new chemistry networks in the future.
> 
> How can you avoid it? You need everything in place to test the code out. Perhaps one alternative is to create _sub-PRs_ by dividing this PR into several chunks and merging them one by one. But there is no way of testing these sub-PRs, and no guarantee that all the chunks when merged would actually work. ISM chemistry in the presence of metals and dust is inherently complex, more so for our science goals that require going to very low metallicities _and_ high densities.
> 

There is a lot of duplicated code in the logic to test the outputs. That can be made simpler, or at least moved to a script that can be called each time. Also, the full reference outputs don't need to be saved in the repo: instead you can just save the final state and/or lines you want to compare.

BenWibking:
The main problem with the size of this PR comes from the 36,000 lines of the full reference outputs. You can instead save just the lines/numbers that you actually compare. Then it should be very manageable. The comparison might be easier to do as a Python script.

unit_test/burn_cell_metal_chem/burn_cell.H
```diff
@@ -0,0 +1,378 @@
...
+      break;
+    case 32:
+      numdens[n - 1] = primary_species_32;
+      break;
+    case 33:
+      numdens[n - 1] = primary_species_33;
+      break;
+    case 34:
+      numdens[n - 1] = primary_species_34;
+      break;
```
BenWibking, line 132: @zingale Is there some template or macro magic we can use to rewrite these very long `switch` statements?

unit_test/burn_cell_metal_chem/burn_cell.H
```diff
@@ -0,0 +1,378 @@
...
+
+    // scale the number densities
+    for (nn = 0; nn < NumSpec; ++nn) {
+      state.xn[nn] *= dd / dd1;
+    }
+
+    // update the number density of electrons due to charge conservation
+    state.xn[2] = -state.xn[5] - state.xn[9] + state.xn[3] + state.xn[6] + state.xn[8] +
+                  2.0 * state.xn[13] + state.xn[11] + state.xn[14] + state.xn[16] + state.xn[21] +
+                  state.xn[24] + state.xn[26] + state.xn[28] + state.xn[29] + state.xn[31];
```
BenWibking, line 262: Can you provide a function that users can call to do this? This has to be done in Quokka as well. This can be hard-coded, as long as it's in a function defined in each network that can be re-used.

EOS/metal_chem/actual_eos.H
```diff
@@ -0,0 +1,423 @@
...
+    if (idx >= 0) {
+        gammas[idx] = eos_rp::species_33_gamma;
+        spmasses[idx] = eos_rp::species_33_mass;
+    }
+
+    idx = network_spec_index(eos_rp::species_34_name);
+    if (idx >= 0) {
+        gammas[idx] = eos_rp::species_34_gamma;
+        spmasses[idx] = eos_rp::species_34_mass;
+    }
```
BenWibking, line 232: It would be nice to replace this with template and/or macro magic if possible.

unit_test/burn_cell_metal_chem/burn_cell.H
zingale: no, but there should be

yut23: Is the loop and switch statement actually needed, or can you do this instead?
```cpp
amrex::Real numdens[NumSpec] = {
    primary_species_1,
    primary_species_2,
    primary_species_3,
    ...,
    primary_species_34
};
numdens[17 - 1] *= metal;
numdens[23 - 1] *= metal;
```
(if you want to allow `NumSpec` < 34, you could fill `amrex::Real primary_species[34]` instead then copy to `numdens`)

EOS/metal_chem/actual_eos.H
yut23: I think a simple macro would work nicely here, along the lines of (untested)
```cpp
#define GET_SPECIES_PARAMS(num) do { \
    idx = network_spec_index(eos_rp::species_## num ##_name); \
    if (idx >= 0) { \
        gammas[idx] = eos_rp::species_## num ##_gamma; \
        spmasses[idx] = eos_rp::species_## num ##_mass; \
    } \
} while (0)

GET_SPECIES_PARAMS(1);
GET_SPECIES_PARAMS(2);
...
GET_SPECIES_PARAMS(34);

#undef GET_SPECIES_PARAMS
```

unit_test/burn_cell_metal_chem/burn_cell.H
zingale: I don't think we want Boost as a dependency

EOS/metal_chem/actual_eos.H
BenWibking: You could even combine, e.g.:
```
#include <boost/preprocessor.hpp>

#define TO_SEQ_ELEM(z, n, data) (n)
#define MAKE_INTEGER_SEQUENCE(n) BOOST_PP_REPEAT(n, TO_SEQ_ELEM, )

MAKE_INTEGER_SEQUENCE(10) // (0) (1) (2) (3) (4) (5) (6) (7) (8) (9)
```

and then:

```
#include <boost/preprocessor/cat.hpp>
#include <boost/preprocessor/seq/for_each.hpp>

#define SEQ (w)(x)(y)(z)

#define MACRO(r, data, elem) (elem, data)

BOOST_PP_SEQ_FOR_EACH(MACRO, _, SEQ) // expands to w_ x_ y_ z_
```

Then you don't even have to unroll `GET_SPECIES_PARAMS` manually.

unit_test/burn_cell_metal_chem/burn_cell.H
BenWibking: I think this should work:
```
#include <boost/preprocessor/cat.hpp>
#include <boost/preprocessor/seq/for_each.hpp>
#include <boost/preprocessor.hpp>

#define TO_SEQ_ELEM(z, n, data) (n)
#define MAKE_INTEGER_SEQUENCE(n) BOOST_PP_REPEAT(n, TO_SEQ_ELEM, )
#define MACRO(r, data, elem) (data, elem)

// this expands to:  primary_species_0, primary_species_1, primary_species_2
BOOST_PP_SEQ_FOR_EACH(MACRO, primary_species_, MAKE_INTEGER_SEQUENCE(3)) 
```

BenWibking: > I don't think we want Boost as a dependency

Ah, ok. maybe we can run the preprocessor on the source and create a custom, specialized version of the macro?

BenWibking: Maybe we can copy-and-paste the implementation for this: https://github.com/HolyBlackCat/macro_sequence_for

yut23: Personally, I find a bunch of lines that only differ by an incrementing number like this is simple enough to scan and check by eye, as long as they're all next to each other.

BenWibking: There's no good reason each network needs to repeat this boilerplate code at all. It could just be autogenerated based on the number of species.

yut23: Another idea I had is to use ParmParse arrays and allow assignments to individual elements, like `unit_test.primary_species(1) = 1e-40`. That would require a bit of work in AMReX, but this seems like a pretty common pattern.

BenWibking: That should already work with `queryArr` if the array elements are separated by spaces, e.g.: `1e-40 1e-40 1e-10` etc.

BenWibking: I agree that using ParmParse to assign individual array elements would be more readable.

EOS/metal_chem/actual_eos.H
psharda: Added the macro. Thanks @yut23 for the suggestion!

unit_test/burn_cell_metal_chem/burn_cell.H
psharda: I created a function in networks that can be called to do this.

unit_test/burn_cell_metal_chem/burn_cell.H
psharda: > That would require a bit of work in AMReX, but this seems like a pretty common pattern.

So this is outside the scope of this PR?

yut23: Yeah, we can come back to that later

.github/workflows/burn_cell_metal_chem.yml
```diff
@@ -0,0 +1,126 @@
...
+          original_input_file="inputs_metal_chem_1"
+          modified_input_file="inputs_metal_chem_modified"
+
+          for Z in "Z=1" "Z=1_z10" "Z=1e-1" "Z=1e-2" "Z=1e-3" "Z=1e-4" "Z=1e-5" "Z=1e-6"; do
+            cp $original_input_file $modified_input_file
+            Z_val=${Z//Z=/}
+
+            if [[ "$Z" == "Z=1_z10" ]]; then
+              # Modify the redshift line for Z=1_z10
+              sed -i 's/network.redshift = 0.0/network.redshift = 10.0/g' $modified_input_file
```
BenWibking, line 75: Instead of modifying the input file, you could just add the new input parameters on the command line. They should override those provided in the input file.

EOS/primordial_chem/actual_eos.H
```diff
@@ -27,88 +27,88 @@ void actual_eos_init ()
...
+        spmasses[idx] = eos_rp::species_13_mass;
     }
 
-    idx = network_spec_index(eos_rp::species_n_name);
+    idx = network_spec_index(eos_rp::species_14_name);
     if (idx >= 0) {
-        gammas[idx] = eos_rp::species_n_gamma;
-        spmasses[idx] = eos_rp::species_n_mass;
+        gammas[idx] = eos_rp::species_14_gamma;
+        spmasses[idx] = eos_rp::species_14_mass;
```
BenWibking, line 111: You could use the macro trick here, too.

networks/metal_chem/actual_rhs.H
```diff
@@ -0,0 +1,4245 @@
...
+        Real rel_t = std::abs((Tdnew - Tdold) / Tdold);
+        Real abs_t = std::abs(Tdnew - Tdold);
+        Tdold = Tdnew;
+
+        if (abs_t < 1e-8 || rel_t < 1e-5) {
+            break;
+        }
+
+        if (iterr > 1000) {
+            throw std::runtime_error("Maximum iterations reached in dust temperature NR-solver. Tgas = " +
```
BenWibking, line 4169: You can't throw exceptions on the GPU.

networks/metal_chem/actual_rhs.H
```diff
@@ -0,0 +1,4245 @@
...
+            } else {
+                return std::make_pair(mid, mid + 1);
+            }
+        }
+        // If the exact range is not found, return the closest valid indices
+        return std::make_pair(high, high + 1);
+    };
+
+    // Find the right x indices using binary search
+    auto [x1_idx, x2_idx] = binary_search(xArray, x);
```
BenWibking, line 4035: Is the Semenov table not equally-spaced in log density and log temperature?

networks/metal_chem/actual_rhs.H
```diff
@@ -0,0 +1,4245 @@
...
+   )
+   : (
+      0
+   )))
+)
+: (
+   0
+)) + 0.00084373771595996178*x4*(1.3806479999999999e-16*X(0) + 1.3806479999999999e-16*X(1) + 1.3806479999999999e-16*X(10) + 1.3806479999999999e-16*X(11) + 1.3806479999999999e-16*X(12) + 1.3806479999999999e-16*X(13) + 1.3806479999999999e-16*X(14) + 1.3806479999999999e-16*X(15) + 1.3806479999999999e-16*X(16) + 1.3806479999999999e-16*X(17) + 1.3806479999999999e-16*X(18) + 1.3806479999999999e-16*X(19) + 1.3806479999999999e-16*X(2) + 1.3806479999999999e-16*X(20) + 1.3806479999999999e-16*X(21) + 1.3806479999999999e-16*X(22) + 1.3806479999999999e-16*X(23) + 1.3806479999999999e-16*X(24) + 1.3806479999999999e-16*X(25) + 1.3806479999999999e-16*X(26) + 1.3806479999999999e-16*X(27) + 1.3806479999999999e-16*X(28) + 1.3806479999999999e-16*X(29) + 1.3806479999999999e-16*X(3) + 1.3806479999999999e-16*X(30) + 1.3806479999999999e-16*X(31) + 1.3806479999999999e-16*X(32) + 1.3806479999999999e-16*X(33) + 1.3806479999999999e-16*X(4) + 1.3806479999999999e-16*X(5) + 1.3806479999999999e-16*X(6) + 1.3806479999999999e-16*X(7) + 1.3806479999999999e-16*X(8) + 1.3806479999999999e-16*X(9))/(std::sqrt(x1)*x35)));
+}
```
BenWibking, line 4010: Up to you, but this file would be a lot easier to read if the cxxcode-generated function were in a separate header file.

EOS/primordial_chem/actual_eos.H
psharda: Yes, but I thought lets do this (and charge balance) in a separate PR focused on the primordial chem network

networks/metal_chem/actual_rhs.H
psharda: no

networks/metal_chem/actual_rhs.H
psharda: Can do it in a future PR

.github/workflows/burn_cell_metal_chem.yml
psharda: yeah, that was the other choice. But they accomplish the same thing anyway.

networks/metal_chem/actual_rhs.H
BenWibking: hmm ok. this has to be done for every rhs evaluation, no? in the future, it might be worth considering resampling the table onto a uniform-log grid for performance.

.github/workflows/burn_cell_metal_chem.yml
BenWibking: Yes, but this method is very hard to parse (pun intended).

networks/metal_chem/actual_rhs.H
psharda: Will using `amrex::Error` like the rest of Microphysics suffice?

networks/metal_chem/actual_rhs.H
psharda: yes.
The opacity has a lot of kinks in particular temperature intervals (see the green curve in the attached plot), so the grid will have to be very dense throughout to maintain constant spacing.
[kappa_compare.pdf](https://github.com/user-attachments/files/16769944/kappa_compare.pdf)

networks/metal_chem/actual_rhs.H
BenWibking: That might kill the whole code. It might be better to trigger the VODE retry mechanism. Not sure how to do that, but @zingale can tell us.

networks/metal_chem/actual_rhs.H
```diff
@@ -0,0 +1,4245 @@
...
+        Real rel_t = std::abs((Tdnew - Tdold) / Tdold);
+        Real abs_t = std::abs(Tdnew - Tdold);
+        Tdold = Tdnew;
+
+        if (abs_t < 1e-8 || rel_t < 1e-5) {
+            break;
+        }
+
+        if (iterr > 1000) {
+            amrex::Error("Maximum iterations reached in dust temperature NR-solver. Tgas = " +
```
BenWibking, line 4169: Still not quite sure about what the right thing to do here is.

BenWibking (approved):
Looks good to me, but I'd still like @zingale to weight in on what the right thing to do about the in-line iterative solve for the dust temperature and how to handle convergence failure there.

networks/metal_chem/actual_rhs.H
zingale: in general, we don't abort in the nets, but instead set a failure for the application code to interpret,
`burn_t.success`

As an aside, if you really need 1000 iterations, maybe NR is not up to the task?  Have you tried the hybrid Powell solver (that's what scipy.fsolve() uses): https://github.com/AMReX-Astro/Microphysics/tree/main/util/hybrj

psharda: We do not need 1000 iterations in general. For the tests, the max iterations taken are  ~unity. If we reach 1000 iterations, then something has probably gone wrong.

I see that the max iterations in the hybrid Powell solver is also set to 1000?

Can this be an issue for future work, and not a part of this PR?

networks/metal_chem/_parameters
```diff
@@ -0,0 +1,30 @@
...
+# cutoff for species mass fractions
+small_x                              real               1.e-100
+# redshift for metal chem (Pop II/I star formation)
+redshift                          real               0e0
+# metallicity for metal chem (Pop II/I star formation)
+metallicity                          real               1e0
+# dust to gas ratio relative to solar
+dust2gas_ratio                 real               1e0
+# starting index for gas-phase species (after the ices)
+idx_gas_species                int                0
```
psharda, line 13: Just FYI @BenWibking we need this newly added parameter in quokka, because we need mass fractions there.

networks/metal_chem/actual_rhs.H
BenWibking: Keeping the Newton solver sounds fine. However, is it possible to set `burn_t.success = 0` here? If it's currently not possible to set the failure condition within actual_rhs, then I'm fine with things as-is in this PR.

psharda: Maybe we can. I can pass the burn_t object to this function, and make it return state.success. So when it is called by `actual_rhs`, state.success would be set to false. Perhaps the rest of the microphysics machinery can then take that into account? 

But I would rather leave this as it is for now, and work on it later when there is more time (after the job season).

PR #1644, opened 2024-08-27
Make VODE_MAXORD a runtime parameter

psharda:


zingale:
this should be a compile-time parameter, not a runtime parameter.  We allocate memory based on this.

PR #1646, opened 2024-08-29
Remove hardcoded charge balance in primordial chemistry

psharda:
We now use a function that does the charge balance for primordial chemistry, similar to that done for metal chemistry in #1642


BenWibking (approved):
🥳

zingale (approved):

PR #1650, opened 2024-09-17
Correctly handle ices in metal chemistry

psharda:
Minor modifications to ensure we handle ices correctly in the metal chemistry network. Also updates the reference solution outputs (minor differences from what we already have).


zingale (approved):

zingale:
you need to update the failing benchmark

PR #1651, opened 2024-09-17
Add cosmic rays to metal chemistry

psharda:
We build on #1642 to add cosmic ray chemistry and heating to the metal chemistry network. Two new tests are added, which use cosmic rays and evolves networks at 1 and 1e-4 Solar metallicity.


psharda:
Right now, I have just added a new RHS file that includes cosmic rays. However, I could also just replace the current actual_rhs.H with this file, since setting the cosmic ray ionization rate to 0 reduces the new RHS to the existing one. But if a user does not want to include cosmic rays, using the new RHS file might take slightly longer to integrate the network (because of the added terms in the ydots and Eint_dot). So is it worth keeping two separate RHS files, or just using the new (more general) one?

psharda:
> Right now, I have just added a new RHS file that includes cosmic rays. However, I could also just replace the current actual_rhs.H with this file, since setting the cosmic ray ionization rate to 0 reduces the new RHS to the existing one. But if a user does not want to include cosmic rays, using the new RHS file might take slightly longer to integrate the network (because of the added terms in the ydots and Eint_dot). So is it worth keeping two separate RHS files, or just using the new (more general) one?

The added time is negligible in one zone tests, but it might not be negiligble in Quokka at high densities and with several AMR levels.

BenWibking:
You could make it a compile-time parameter, e.g. `enable_cosmic_rays`?

psharda:
Decided to just have one RHS for now, can come back to having multiple later if needed for Quokka.

psharda:
omukai2005_i
Attaching this plot for future reference. Solid curves are with cosmic rays included (for a fixed CR ionization rate of 3e-17 per s), dashed curves are without cosmic rays. As expected, cosmic ays heat the gas at low density, high metallicity, and cool the gas at low density, low metallicity.

BenWibking (approved):
Looks good to me.

PR #1660, opened 2024-10-24
switch CNO_He_burn and subch_simple to derive reverse rates

zingale:
this makes them consistent with the other nets


zingale:
going to redo this one now that we've reorganized the nets

PR #1686, opened 2024-12-10
update HIP/CUDA dependences to include sparse libraries

zingale:
these are needed by AMReX


yut23 (approved):
For future reference, these scripts are copied from https://github.com/AMReX-Codes/amrex/tree/9643da4522f955767ab78bc03abe206d238f8949/.github/workflows/dependencies

PR #1688, opened 2024-12-13
Implement Debye Huckel Screening

SamG-01:
Implements weak screening as per equation A1 of Chugunov 2009 (see below). Also adds an optional check against `debye_huckel` and computes the full screening factor if the check exceeds a supplied threshold, which can significantly improve runtimes (disabled by default).

image


screening/screen.H
```diff
@@ -152,6 +153,33 @@ fill_plasma_state(plasma_state_t<number_t>& state, const number_t& temp,
...
+    // scn_fac = screening factors for A and Z
+
+    amrex::Real z1z2 = scn_fac.z1 * scn_fac.z2;
+    amrex::Real z_ratio = state.z2bar / state.zbar;
+
+    // Gamma_e from eq. 6
+    number_t Gamma_e = state.gamma_e_fac / state.temp;
+
+    // eq. A1
+    number_t h_DH = z1z2 * gcem::sqrt(3 * amrex::Math::powi<3>(Gamma_e) * z_ratio);
```
yut23, line 175:
```suggestion
    number_t h_DH = z1z2 * admath::sqrt(3 * amrex::Math::powi<3>(Gamma_e) * z_ratio);
```

screening/screen.H
```diff
@@ -618,13 +646,24 @@ number_t chabrier1998 (const plasma_state_t<number_t>& state,
 template <typename number_t>
 AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
 number_t actual_screen(const plasma_state_t<number_t>& state,
-                     const scrn::screen_factors_t& scn_fac)
+                     const scrn::screen_factors_t& scn_fac,
+                     const bool& skip_check = false,
+                     const number_t& skip_threshold = 1.01_rt)
```
yut23, line 651: Passing primitive types by value is cheaper, and `skip_threshold` doesn't need to be a dual number when we're calculating derivatives (same for the other `actual_screen()` overload below).
```suggestion
                     const bool skip_check = false,
                     const amrex::Real skip_threshold = 1.01_rt)
```

screening/screen.H
```diff
@@ -618,13 +646,24 @@ number_t chabrier1998 (const plasma_state_t<number_t>& state,
...
 {
-    number_t scor;
+    number_t scor = 1.0_rt;
 #if SCREEN_METHOD == SCREEN_METHOD_null
     // null screening
     amrex::ignore_unused(state, scn_fac);
-    scor = 1.0_rt;
+    return scor;
+#endif
+    if skip_check {
```
yut23, line 659:
```suggestion
    if (skip_check) {
```

screening/screen.H
```diff
@@ -641,10 +680,12 @@ template <typename number_t>
 AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
 void actual_screen(const plasma_state_t<number_t>& state,
                    const scrn::screen_factors_t& scn_fac,
+                   const bool& skip_check = false,
+                   const number_t& skip_threshold = 1.01_rt,
```
yut23, line 684: These parameters with defaults need to go after `scor` and `scordt`

screening/screen.H
```diff
@@ -618,13 +646,24 @@ number_t chabrier1998 (const plasma_state_t<number_t>& state,
 template <typename number_t>
 AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
 number_t actual_screen(const plasma_state_t<number_t>& state,
-                     const scrn::screen_factors_t& scn_fac)
+                     const scrn::screen_factors_t& scn_fac,
+                     const bool skip_check = false,
```
yut23, line 650: Glancing over the code again, this argument name is a bit confusing. My first thought was that it should skip the check if false, so maybe rename it to `enable_skip_check` or something similar?

screening/screen.H
```diff
@@ -618,13 +646,24 @@ number_t chabrier1998 (const plasma_state_t<number_t>& state,
 template <typename number_t>
 AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
 number_t actual_screen(const plasma_state_t<number_t>& state,
-                     const scrn::screen_factors_t& scn_fac)
+                     const scrn::screen_factors_t& scn_fac,
+                     const bool enable_skip_check = screening_rp::enable_debye_huckel_skip,
```
yut23, line 650: If this is a runtime parameter, it's simpler to check it directly inside the function body rather than adding an argument with a different name. You can probably go ahead and do the same for `skip_threshold` too.

yut23 (approved):

zingale (approved):

SamG-01:
> [...] computes the full screening factor if the check exceeds a supplied threshold, which can significantly improve runtimes (disabled by default).
> 

To quantify these runtime improvements, below is a benchmark of the `subchandra` simulation in `Castro`, with and without `screening.enable_debye_huckel_skip`, for several values of `castro.react_rho_min`. The stop time is 0.01 s, and `screening.debye_huckel_skip_threshold` is kept at its default of 1.01.

output

PR #1690, opened 2025-01-04
Update unit test docs

zingale:
The `burn_cell` docs were very out of date.  This makes them accurate and also
adds a new plot script for looking at the output.  It also removes the unused prefix
from the test's parameters and removes an old `matplotlibrc` file.


Docs/source/one_zone_tests.rst
```diff
@@ -9,97 +9,173 @@ on a single zone to inspect the output directly.
...
-determine the needed ODE tolerances.
+determine the needed ODE tolerances.  This is designed to work
+with the Strang-split integration wrappers.  The system that is evolved
+has the form:
+
+.. math::
+
+   \begin{align*}
+      \frac{dX_k}{dt} &= \dot{\omega}_k(\rho, X_k, T) \\
+      \frac{de}{dt} &= f(\rho, X_k, T)
```
zhichen3, line 25: I think we want to use `epsilon` notation now.

Docs/source/one_zone_tests.rst
```diff
@@ -9,97 +9,173 @@ on a single zone to inspect the output directly.
...
+   \end{align*}
+
+with density held constant and the temperature found via the equation of state,
+$T = T(\rho, X_k, e)$.
+
+
+.. note::
+
+   Since the energy evolves due to the heat release (or loss)
+   from reactions, the temperature will change during the burn.
```
zhichen3, line 35: maybe we should also mention temperature evolution can be optionally disabled via `integrator.call_eos_in_rhs=0` to offer more flexibility for users to explore the difference.

zingale: updated

Docs/source/one_zone_tests.rst
zingale: updated

Issue #1680, opened 2024-12-01
document neutrinos


zingale:
The neutrino loss terms are not documented.

PR #1692, opened 2025-01-05
Add neutrino docs

zingale:
this also makes the notation for the energy generation rate consistent across pages


zingale:
closes #1680

Issue #1693, opened 2025-01-05
breakout EOS references aux[1]


zingale:
the difference between `gamma_law` and `breakout` is simply the computation of mu,

```
    // Calculate mu. This is the only difference between
    // this EOS and gamma_law.
    state.mu = 1.0_rt / state.aux[1];
```

but we don't seem to have any check whether `aux` has Abar as component `[1]`.  This is used in Castro
by the `RadBreakout` which uses `general_null` and `breakout.net`, which defines

```
# auxiliary variables
__aux_Ye
__aux_invmu
```

so we should add a check to the `eos_init` that makes sure that this aux variable is present.

PR #1694, opened 2025-01-05
make sure the breakout EOS is used with a aux that is 1/mu defined

zingale:
now we check this in the eos_init and use the proper aux index when accessing aux[]


zingale:
closes #1693

zhichen3 (approved):

PR #1696, opened 2025-01-05
[WIP] try to get doxygen working

zingale:
the big issue seems to be that we have lots of files with the same name (like `actual_eos.H`, so we need Doxygen to understand how to deal with those.

### DIFF
--- a/sphinx_docs/Doxyfile
+++ b/sphinx_docs/Doxyfile
@@ -425,7 +425,7 @@ LOOKUP_CACHE_SIZE      = 0
 # normally produced when WARNINGS is set to YES.
 # The default value is: NO.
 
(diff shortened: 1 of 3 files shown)


sphinx_docs/Doxyfile
```diff
@@ -839,7 +839,7 @@ EXCLUDE_SYMLINKS       = NO
 # Note that the wildcards are matched against the file with absolute path, so to
 # exclude all test directories for example use the pattern */test/*
 
-EXCLUDE_PATTERNS       =
+EXCLUDE_PATTERNS       = util/autodiff/* util/gcem/* util/skynet/* */tmp_build_dir/*
```
yut23, line 842: These may need to have a leading `*/`

PR #1699, opened 2025-01-05
fix the CJ detonation utility

zingale:


zingale:
This works again with `aprox13`.  I've updated the `README.md` a bit.
Here's the plot it creates:

cj_det

maxpkatz (approved):

PR #1701, opened 2025-01-06
[WIP] regenerate the networks with the latest pynucastro

zingale:
this updates binding energies with more precision, fixes partition functions with ROCm/HIP, and adds the ability to compile out neutrino losses.


zingale:
going to regenerate once the dust settles

PR #1714, opened 2025-01-13
remove runtime_parameters.rst from git

zingale:
it is automatically generated


.gitignore
```diff
@@ -12,6 +12,7 @@
 *.ind
 Docs/*/*.pdf
 Docs/*/*.ps
+Docs/source/runtime_parameters.rst
```
zhichen3, line 15: isn't that already there under # sphinx

zingale: huh... indeed.

so how did this file get added then?

zingale: ooohhh...

it happened in #1689 when I renamed the directory.

I'll update this PR just to remove the file

zingale: okay... this just removes the file that snuck in now.

zhichen3 (approved):

PR #1717, opened 2025-01-14
remove some old Fortran stuff from the build system

zingale:


zhichen3 (commented):
not sure if it needs to be removed, but line 86 also mention `USE_COMPILE_WITH_F2PY`

zhichen3 (approved):

PR #1725, opened 2025-01-27
update pynucastro networks for pynucastro 2.5.0

zingale:
in particular, this will make the physical constants more consistent, with better binding energies.  It also will ensure all nets work with ROCm by fixing the partition function initialization.


zingale:
going to close and redo this one

PR #1729, opened 2025-01-28
rename test_nse_net -> nse_net_cell; test_part_func -> part_func_cell

zingale:
this is consistent with the other naming conventions that these are one-zone tests


zhichen3:
we should change the heading in the README for nse_net_cell too.

zingale:
readme updated

PR #1732, opened 2025-01-28
fix some markdown in CHANGES.md and add some more PR refs

zingale:
this also now makes the changelog part of the docs with links to the PRs.

### DIFF
--- a/Docs/source/conf.py
+++ b/Docs/source/conf.py
@@ -107,7 +107,7 @@ language = "en"
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
 # This patterns also effect to html_static_path and html_extra_path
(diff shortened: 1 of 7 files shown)


Docs/source/conf.py
```diff
@@ -107,7 +107,7 @@ def get_version():
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
 # This patterns also effect to html_static_path and html_extra_path
-exclude_patterns = []
+exclude_patterns = ["changelog.md"]
```
yut23, line 110: Should this also be added to the gitignore?

zingale: yeah, wouldn't hurt.  I'll do that

yut23 (approved):

PR #1749, opened 2025-02-11
[WIP] centralize some runtime parameters and their docs

zingale:


Docs/source/unit_test_runtime_parameters.rst
```diff
@@ -0,0 +1,58 @@
...
+.. _sec:defining_unit_test_composition:
+
+Defining composition
+====================
+
+Most of the unit tests require a composition to be defined (for the
+initial mass-fractions, $X_k$).  There are a few ways this can be done
+(depending on the test).
+
+* One-zone (``*_cell``) tests usually do one of:
```
zhichen3, line 28: I think it would aid the reader if we reference one-zone test and comprehensive test to their corresponding section when we mention them here

unit_test/part_func_cell/_parameters
```diff
@@ -1,7 +1,5 @@
 @namespace: unit_test
 
-small_temp    real       1.e5
-small_dens    real       1.e5
-
+density       real    1.e9
```
zhichen3, line 3: where is density used in this unit test?

unit_test/burn_cell_metal_chem/_parameters
```diff
@@ -3,8 +3,8 @@
 run_prefix    string  "burn_cell_metal_chem"
 
 # floor values of temperature and density
-small_temp    real       1.e1
-small_dens    real       1.e-30
+small_temp    real       1.e1           100
```
zhichen3, line 6: what does `100` do in the end?

zingale: that sets the priority, meaning that this value will be used instead of the one from `unit_test/_parameters`

zingale: that's described here:
https://amrex-astro.github.io/Microphysics/docs/rp_intro.html

unit_test/part_func_cell/_parameters
zingale: it's not.  I just removed it.   I feel like I did that once already in a different pr...

Docs/source/unit_test_runtime_parameters.rst
zingale: added

unit_test/burn_cell_metal_chem/_parameters
zhichen3: ah okay. Didn't know that before.

zhichen3 (approved):

PR #1757, opened 2025-02-14
explicitly set small_{temp,dens} for some tests

zingale:


zingale:
this restores some values from before #1749

zingale:
this fixes some test suite failures

zhichen3 (approved):

PR #1764, opened 2025-02-22
[WIP] a new He network with 38 nuclei

zingale:
this eliminates the endpoint at Na22 and includes Zn60


zingale:
this network doesn't really break the odd-Z to even-Z link.  I can do better now with modified rates

Issue #1771, opened 2025-03-03
Add support for a single-step Backward Euler without error control


chongchonghe:
It would be helpful to allow forcing single-step Backward Euler integration without worrying about errors, by setting some parameter, say `integrator.atol_enuc = -1` . There are situations where we want to control relative error by setting the hydro/radiation timestep, and in a reaction network we simply want a backward Euler solve and don't care about error. 

Context: this is part of the effort to couple multi-group radiation hydrodynamics with thermochemistry in Quokka. 

@zingale

PR #1773, opened 2025-03-05
Allow single-step backward Euler integrator

chongchonghe:
This allows the user to use a single-step Backward Euler integrator, ignoring its accuracy. There are situations where we control relative error by setting the hydro/radiation timestep, and in a reaction network we simply want a backward Euler solve and don't care about error.

To force the BackwardEuler integrator to do a single-step integration, set `integrator.do_single_step = 1`. 

Verified with the `burn_cell` unit test:

```bash
cd unit_test/burn_cell
make INTEGRATOR_DIR=BackwardEuler
echo "integrator.do_single_step = 1" >> inputs_aprox13
./main3d.gnu.ex inputs_aprox13 
```

Context: this is part of the effort to couple multi-group radiation hydrodynamics with thermochemistry in Quokka.

Addresses #1771

### DIFF
--- a/Docs/source/ode_integrators.rst
+++ b/Docs/source/ode_integrators.rst
@@ -27,9 +27,9 @@ Presently, allowed integrators are:
 * ``BackwardEuler``: an implicit first-order accurate backward-Euler
   method.  An error estimate is done by taking 2 half steps and
   comparing to a single full step.  This error is then used to control
-  the timestep by using the local truncation error scaling. Optionally, the 
-  user can disable error estimation and force a single-step backward-Euler 
-  integration by setting `integrator.atol_enuc = -1`.
+  the timestep by using the local truncation error scaling. Optionally, the
+  user can disable error estimation and force a single-step backward-Euler
+  integration by setting `integrator.do_single_step = 1`.
 
 * ``ForwardEuler``: an explicit first-order forward-Euler method.  This is
   meant for testing purposes only.  No Jacobian is needed.
--- a/Docs/source/ode_integrators.rst
+++ b/Docs/source/ode_integrators.rst
@@ -27,7 +27,9 @@ Presently, allowed integrators are:
 * ``BackwardEuler``: an implicit first-order accurate backward-Euler
   method.  An error estimate is done by taking 2 half steps and
   comparing to a single full step.  This error is then used to control
-  the timestep by using the local truncation error scaling.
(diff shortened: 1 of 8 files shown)


zingale:
can you also add a note to the docs here:
https://github.com/AMReX-Astro/Microphysics/blob/main/Docs/source/ode_integrators.rst

zingale:
you also need to fix the style CI failures -- it looks like you used tabs and have trailing whitespace.

zingale:
also, rather than making `atol_enuc` special (e.g., why is this set to -1 and not the others?) it might be better to simply define a new parameter in `Microphysics/integration/BackwardEuler/_parameters`, like `do_single_step` that defaults to `0`, but if set to `1` does what you want.

I think that would be cleaner.

chongchonghe:
I've added a `integrator.do_single_step` parameter to toggle single-step EU. I tested it on my computer and it worked file.

zingale:
you can test it with GNU make -- we don't use cmake.  Just do:

```
make INTEGRATOR_DIR=BackwardEuler
```

in the test directory.

chongchonghe:
> you can test it with GNU make -- we don't use cmake. Just do:
> 
> ```
> make INTEGRATOR_DIR=BackwardEuler
> ```
> 
> in the test directory.

Thanks, that works. I kept the changes minimal and it works just fine. Would you merge it into the main branch if everything is okay? @zingale

Docs/source/ode_integrators.rst
```diff
@@ -27,7 +27,9 @@ Presently, allowed integrators are:
 * ``BackwardEuler``: an implicit first-order accurate backward-Euler
   method.  An error estimate is done by taking 2 half steps and
   comparing to a single full step.  This error is then used to control
-  the timestep by using the local truncation error scaling.
+  the timestep by using the local truncation error scaling. Optionally, the 
+  user can disable error estimation and force a single-step backward-Euler 
+  integration by setting `integrator.atol_enuc = -1`.
```
zingale, line 32: this needs to be updated now

zingale:
there are also still CI style failures.

Docs/source/ode_integrators.rst
chongchonghe: Done.

chongchonghe:
> there are also still CI style failures.

Fixed.

PR #1777, opened 2025-04-08
Adding new alternate neutrino cooling

khanakbhargava:
This is the first draft of a neutrino cooling routine using the kippenhahn et al. approximations. I have used autodiff. I will compare these with itoh et al and give the options to choose between the two.


khanakbhargava:
The pair annihilation approximations only account for T9 < 1 and T9 > 3. We need to find a way to blend the two. The book does not say anything about the temp in between.

yut23:
Quick and dirty plot of the pair annihilation asymptotics ([relevant page from archive.org](https://archive.org/details/StellarStructureAndEvolutionKippenhahnWeigert/page/n90/mode/1up)):
kipp_pair_annihilation

zingale:
see page 2 of Stans notes here:
https://www.ucolick.org/~woosley/ay220-15/lectures/lecture11.4x.pdf

zingale:
Also, Clayton Eq 3-302 has the expression for T_9 < 1 as 4.9e18 -- I wonder if K&W have a typo and dropped the "1" in the "18" exponent?

khanakbhargava:
docs build check is failing right now, but I fixed the rest issues.

khanakbhargava:
kipp
itoh

This plot shows the neutrino cooling rate w.r.t temperature. I need to check what is happening at the lower temperatures.

zingale:
this source discusses some new approximations:
https://www.sciencedirect.com/science/article/pii/S0550321303001512#BIB023

neutrinos/kipp.H
```diff
@@ -0,0 +1,103 @@
...
+    // eq. 18.86
+    number_t T8 = T9 * 10.0e0_rt;
+    return 0.76e0_rt * (amrex::Math::powi<2>(zbar)/abar) * amrex::Math::powi<6>(T8);
+}
+
+AMREX_GPU_HOST_DEVICE AMREX_INLINE
+void kipp(const amrex::Real& temp, const amrex::Real& rho, const amrex::Real& abar,
+    const amrex::Real& zbar, amrex::Real& snu, amrex::Real& dsnudt, amrex::Real& dsnudrho, amrex::Real& dsnudz, amrex::Real& dsnuda){
+
+    amrex::Real T9 = temp * 1.0e-9_rt;
```
yut23, line 73: This makes `dsnudt` the derivative with respect to T9, which doesn't match sneut5

neutrinos/kipp.H
```diff
@@ -0,0 +1,103 @@
...
+    number_t mu_e = abar / zbar;
+
+    return ep_1 + ep_2 / (mu_e + rho_bar);
+}
+
+template<typename number_t>
+AMREX_GPU_HOST_DEVICE AMREX_INLINE
+// plasmaneutrinos
+number_t kip_plas(const number_t& rho, const number_t& T9, const number_t& abar, const number_t& zbar){
+    number_t n_e = rho / (C::m_u);
```
yut23, line 46: Should this be $\rho/(\mu_e m_u)$?

neutrinos/kipp.H
```diff
@@ -0,0 +1,103 @@
...
+    // assuming non-degeneracy
+    // 18.83
+    number_t w0 = admath::sqrt((4.0e0_rt * M_PI * amrex::Math::powi<2>(C::q_e) * n_e) / C::m_e);
+    number_t gamma = (C::hbar * w0) / (C::k_B * T9);
+    number_t lambda = (C::k_B * T9) / (C::m_e * amrex::Math::powi<2>(C::c_light));
```
yut23, line 51: These should use T, not T9

khanakbhargava:
kipp_main

I think we have better agreement now.

yut23:
Oh right, you need to wrap the argument to `amrex::Math::powi()` with `autodiff::eval()`. I should add that to the docs.

neutrinos/kipp.H
```diff
@@ -0,0 +1,177 @@
...
+    // 18.83 replaced by aprox in woosley lecture notes
+    number_t w0 = admath::sqrt((4.0e0_rt * M_PI * amrex::Math::powi<2>(C::q_e) * n_e) / C::m_e);
+    number_t gamma = (C::hbar * w0) / (C::k_B * temp);
+    number_t lambda = (C::k_B * temp) / (C::m_e * amrex::Math::powi<2>(C::c_light));
+    // 18.85
+   if (gamma <= 1.0e0_rt) {
+       // return (7.4e21_rt * amrex::Math::powi<6>(C::hbar * w0/(C::m_e * amrex::Math::powi<2>(C::c_light)))
+       //         * amrex::Math::powi<3>(lambda))/rho;
+        auto epsilon_p = (C::hbar * w0) / (C::m_e * C::c_light * C::c_light);
+        return (7.4e21_rt * admath::pow(epsilon_p, 6.0_rt) * admath::pow(lambda, 3.0_rt)) / rho;
```
yut23, line 63: When assigning an intermediate expression to a value, either the destination type should be `number_t` or the expression should be wrapped with `autodiff::eval()`, to force the expression template to be evaluated.
```suggestion
        number_t epsilon_p = (C::hbar * w0) / (C::m_e * C::c_light * C::c_light);
        return (7.4e21_rt * amrex::Math::powi<6>(epsilon_p) * amrex::Math::powi<3>(lambda)) / rho;
```

neutrinos/neutrino.H
```diff
@@ -0,0 +1,39 @@
...
+number_t neutrino_cooling(const number_t& temp, const amrex::Real dens,
+                          const number_t& abar, const number_t& zbar) {
+
+#if NEUTRINO_METHOD == NEUTRINO_METHOD_kipp
+    amrex::Real pair, phot, plas, brem;
+    return kipp(temp, dens, abar, zbar, pair, phot, plas, brem);
+
+#elif NEUTRINO_METHOD == NEUTRINO_METHOD_sneut5
+    auto sf = get_sneut_factors(dens, temp, abar, zbar);
+    return nu_pair(sf) + nu_photo(sf) + nu_plasma(sf) + nu_brem(sf);
```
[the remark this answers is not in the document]
yut23: And also include `nu_recomb` if applicable

zingale:
this just needs to have documentation added before merging.

unit_test/test_neutrino_cooling/neutrino_util.cpp
```diff
@@ -76,16 +78,31 @@ void neut_test_C(const Box& bx,
...
+    Real pair{};
+    Real phot{};
+    Real plas{};
+    Real brem{};
 
     constexpr int do_derivatives{1};
 
+    #if NEUTRINO_METHOD == NEUTRINO_METHOD_kipp
+    kipp<do_derivatives>(temp_zone, dens_zone, abar, zbar,
+                         snu, dsnudt, dsnudd, dsnuda, dsnudz,
```
zingale, line 90: here you have the order as `dsnuda, dsnudz`, but in `kipp.H`, line 119, the order is swapped.

You should make the argument ordering consistent with `sneut5` to prevent this type of error.

neutrinos/kipp.H
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+}
+
+template<typename number_t>
+AMREX_GPU_HOST_DEVICE AMREX_INLINE
+// bremsstrahlung neutrinos
+number_t kip_brem(const number_t& temp, const amrex::Real& rho,
+                  const number_t& abar, const number_t& zbar){
+    // eq. 18.86
```
zingale, line 74: add:

```
    amrex::ignore_unused(rho);
```

unit_test/test_neutrino_cooling/neutrino_util.cpp
```diff
@@ -76,16 +78,31 @@ void neut_test_C(const Box& bx,
...
     Real dsnuda;
     Real dsnudz;
+    Real pair{};
+    Real phot{};
+    Real plas{};
+    Real brem{};
 
     constexpr int do_derivatives{1};
 
+    #if NEUTRINO_METHOD == NEUTRINO_METHOD_kipp
```
zingale, line 88: don't indent the preprocess directives (it's legal, but inconsistent with how we usually code these)

Make.Microphysics_extern
```diff
@@ -44,6 +44,15 @@ else
   $(error Invalid value for SCREEN_METHOD)
 endif
 
+NEUTRINO_METHOD ?= kipp
```
zingale, line 47: why are you making the default `kipp`?  I think we want the default to be `sneut5` so people opt-in to this alternate method.

networks/rhs.H
```diff
@@ -1325,8 +1326,19 @@ void rhs (burn_t& burn_state, amrex::Array1D<amrex::Real, 1, nrhs>& ydot)
 #ifdef NEUTRINOS
     constexpr int do_derivatives{0};
     amrex::Real sneut, dsneutdt, dsneutdd, dsnuda, dsnudz;
-    sneut5<do_derivatives>(burn_state.T, burn_state.rho, burn_state.abar, burn_state.zbar,
-                           sneut, dsneutdt, dsneutdd, dsnuda, dsnudz);
+    amrex::Real pair, phot, plas, brem;
+
+#if NEUTRINO_METHOD == NEUTRINO_METHOD_kipp
```
zingale, line 1331: I think it's better to just have the different networks call a wrapper that has this logic, instead of updating all the calls like this.  That way if we add another neutrino cooling implementation in the future, we don't need to update them all again.

khanakbhargava: Oh I already do have that, I forgot about it. I will use it. it is in `neutrinos/neutrino.H`

PR #1780, opened 2025-04-09
disable the inline hack with HIP

zingale:
we can how toggle it via a build parameter


zingale:
I've tested whether we still need these compilation flags on Frontier with `flame_wave`  and the `he_burn/he_burn_22a` network, and it seems like we are fine without them (when using ROCm 6.3.1).

zingale:
the 3D massive star also works fine on Frontier without these flags

PR #1782, opened 2025-04-10
[WIP] start of a more complete he burning net

zingale:


zingale:
closing this as none of these were really great in the end.  I will start on a new set with the latest pynucastro.

PR #1786, opened 2025-05-07
Adding option to choose recombination in neutrino cooling

khanakbhargava:
While working on adding Kippenhahn approximations using autodiff, we realized that mostly everyone does not include recombination when calculating the cooling rates since it does not have much effect on it. Here, I have added the option to choose whether or not we want to include the recombination term. The plots show the output of the unit test with and without recombination. 
recomb
no_recomb

### DIFF
--- a/Microphysics/neutrinos/sneut5.H
+++ b/Microphysics/neutrinos/sneut5.H
@@ -1,0 +1,1573 @@
+#ifndef SNEUT5_H
+#define SNEUT5_H
+
+#include <AMReX_REAL.H>
+#include <AMReX_Array.H>
+#include <AMReX_Math.H>
+
+using namespace amrex::literals;
+
+namespace nu_constants {
+
+    // numerical constants
+    constexpr amrex::Real fac1   = 5.0e0_rt * M_PI / 3.0e0_rt;
+    constexpr amrex::Real fac2   = 10.0e0_rt * M_PI;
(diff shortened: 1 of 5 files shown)


zingale:
you should make `include_recomb` a runtime parameter, by creating an `_parameters` file in `Microphysics/neutrinos/` and adding it to a new neutrino_cooling namespace.  Look at the format from one of the other modules.

Also you should not change the default -- keep it on by default.  Later in a separate PR, we can decide to change the default once we understand this better.

khanakbhargava:
Made the changes

Microphysics/neutrinos/sneut5.H
```diff
@@ -0,0 +1,1573 @@
...
+#include <AMReX_REAL.H>
+#include <AMReX_Array.H>
+#include <AMReX_Math.H>
+
+using namespace amrex::literals;
+
+namespace nu_constants {
+
+    // numerical constants
+    constexpr amrex::Real fac1   = 5.0e0_rt * M_PI / 3.0e0_rt;
```
zingale, line 13: why are all these lines different?  I don't see what changed from what's in there now

yut23: These files are nested under a new `Microphysics/` directory

khanakbhargava:
I will do another PR. My directories were mixed up.

PR #1788, opened 2025-05-14
He burn modified

zingale:


networks/he-burn/ase/inputs.burn_cell.VODE
zhichen3, line 1: I think we removed these input files.

zhichen3 (approved):

Issue #82, opened 2017-12-20
test_eos does not work for all networks


maximumcats:
test_eos explicitly uses "solar" metallicity, and so it will not work for networks that don't have ih1 or ihe4 defined. We should generalize this so any net can use it.

zingale:
do you have an idea in mind on how to generalize this?

I can just make the composition evenly distributed through the different species.

maxpkatz:
Yes, evenly distributed is the easiest, I am fine with that.

PR #1792, opened 2025-05-21
update test_eos to use composition runtime parameters

zingale:
this allows it to work with any network, not just one with H1 and He4


zingale:
closes #82

yut23 (approved):

PR #1798, opened 2025-05-29
add a larger nova network

zingale:
this includes more CNO breakout as well as Li and Be links


zingale:
this is ready for review

zhichen3 (approved):

Issue #1655, opened 2024-09-20
update physical constants and drop "legacy' from the pynucastro nets


zingale:
We should update the non-legacy constants in `fundamental_constants.H` to the same version used by SciPy (since that's where pynucastro gets constants).  We should also switch the pynucastro nets from using the legacy constants to using the new values.

PR #1802, opened 2025-06-06
update constant module to use scipy value and remove using legacy constants

zhichen3:
added a script to auto generate scipy constant value. Address issue #1655.

### DIFF
--- a/constants/write_fundamental_constants.py
+++ b/constants/write_fundamental_constants.py
@@ -1,4 +1,26 @@
+#!/usr/bin/env python3
 
+import os
+import warnings
+import math
+from scipy import constants
+
+'''
+This scripts writes out fundamental_constants.H
+where most constants are from scipy.constants
+'''
+
+# Find PATH to write the file
+MICROPHYSICS_HOME = os.environ.get("MICROPHYSICS_HOME")
+if MICROPHYSICS_HOME is None:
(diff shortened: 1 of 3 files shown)


zingale:
I think that if we update G here, then the initial models that do self-gravity would need to be regenerated, since Castro would pick up this G.  This is fine, but we should coordinate this carefully.

zhichen3:
> I think that if we update G here, then the initial models that do self-gravity would need to be regenerated, since Castro would pick up this G. This is fine, but we should coordinate this carefully.

Right, the old G is not updated. Only the new G is updated, and that's in comment. So its not being used actually

zhichen3:
I also updated the `burn_cell_metal_chem` unit test script to get the "correct" line number. Previously the test actually results in some integration failure, to see that grep for `fail` in `test_*.out`. With this pr, we no longer see those failure. So the reference line number changed, resulting in comparison between different quantities. Here I updated it and the test passes.

zingale:
tests:
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2025-06-07-001/index.html

zingale:
I think we will still need to update any Castro initial models that use gamma_law, since the constants are changing.

zingale:
oh indeed.  I thought it would be more.  I'll review this today.

zingale (approved):

PR #1812, opened 2025-06-19
update nse_sdc_burn for NSE_NET

zhichen3:
Addresses the following:
1) Fix the issue that abar wasn't calculated correctly. Needed to take the reciprocal in the end.
2) Switch to use central difference rather than 5-stencil to compute dabardT. Convergence test shows that central difference is sufficient for 2nd order.
3) Construct the reactive source term for rho e directly, instead of evolving for tau and do a finite difference to get the derivative.

### DIFF
--- a/Docs/source/nse_net.rst
+++ b/Docs/source/nse_net.rst
@@ -93,10 +93,10 @@ Here we give outline on how the reactive update is done with NSE.
     * Now evolve $\rho$, $\rho e$, and $\rho Y_e$ to midpoint in time:
 
       .. math::
-         \Uc^{\prime,n+1/2} = \Uc^{\prime,n} + \frac{\Delta t}{2} \left([\Advs{\Uc^\prime}]^{n} + [\Rb(\Uc^\prime)]^{n}\right)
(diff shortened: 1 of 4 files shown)


zhichen3:
Now with detonation:

Running with direct integration:
image

Running with this PR but without modifying how we calculate drhoedt:
image

Completely running with this PR:
image

With this PR we see that the temperature profile matches with direct integration exactly. We only see energy generation from zones that are not in NSE, i.e. just regular burning. And behind the shock is in NSE.
Since we calculate `drhoedt` directly from `Ydot_weak`, `e_nuc` contribution is identically zero for network without weak rates, and only contributing factors are from advection terms and neutrinos. I think it is a reasonable thing since if we're in NSE, then there is no energy generation from strong reactions? 

With this PR, I'm able to run detonation at a similar runtime compared to direct integration now. Before it was running substantially slower.

zingale:
this is really cool.  Do you understand why we don';t need to do the same finite-difference estimate of the update as we do with the tabular NSE here?

zingale (approved):

PR #1828, opened 2025-06-28
update author list

zingale:
and sync the zenodo file


zingale:
This now has everyone who confirmed that they want to be part of the paper, and the zenodo file is synced with the author list.

PR #1831, opened 2025-07-01
add a CI for validating the Zenodo JSON

zingale:


zingale:
I tested this by intentionally breaking the JSON and the test failed.  So this seems to work.

zhichen3 (approved):

PR #1837, opened 2025-07-12
make the pynucastro networks importable

zingale:
this will allow us to use the net back in pynucastro to plot the network flow from a simulation


zingale:
this updates all of the non-He burning networks.

PR #1840, opened 2025-07-21
Neutrino cooling

khanakbhargava:


zingale:
can you revert the changes to just the pynucastro nets?  We will regenerate these in a later PR, once pynucastro 2.7.0 is released, since there are a lot of changes (some pending) in pynucastro that are not reflected here yet.  It's fine that those nets will still hardcode sneut5 for now.

zingale:
regression tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2025-07-28-001/index.html

the neutrino test there fails because of the renaming of the plotfile, but there are no diffs in a manual fcompare

zingale:
no, I'll handle it after it is merged

neutrinos/Make.package
```diff
@@ -1 +1,2 @@
-CEXE_headers += sneut5.H
+CEXE_headers += neutrino.H
+CEXE_headers += sneut5.H kipp.H
```
zingale, line 2: can you put these each on a separate line
this is probably fine as-is, but we never do this anywhere else (as far as I know)

neutrinos/sneut5.H
```diff
@@ -911,4 +919,16 @@ void sneut5(const amrex::Real temp, const amrex::Real den,
 
 }
 
+template <int do_derivatives>
+AMREX_GPU_HOST_DEVICE AMREX_INLINE
+void sneut5(const amrex::Real temp, const amrex::Real den,
```
zingale, line 924: why do we need this additional interface? is this because of the nets that don't get the individual components?

khanakbhargava: We don't need this anymore. This was before we added a wrapper. I have updated it.

khanakbhargava: Without it the networks were giving errors so I can remove it since we will be updating the networks. 

```
../../networks/partition_test/actual_rhs.H:259:27: error: no matching function for call to ‘sneut5<do_derivatives>(amrex::Real&, amrex::Real&, amrex::Real&, amrex::Real&, amrex::Real&, amrex::Real&, amrex::Real&, amrex::Real&, amrex::Real&)’
  259 |     sneut5<do_derivatives>(state.T, state.rho, state.abar, state.zbar, sneut, dsneutdt, dsneutdd, dsnuda, dsnudz);
```

zingale (approved):

PR #1849, opened 2025-09-19
Regenerate nets partition functions

zingale:


zingale:
This captures the bug fix here:
https://github.com/pynucastro/pynucastro/pull/1100

zhichen3 (approved):

PR #1850, opened 2025-09-24
fix subtle bug in he-burn derived rates

zingale:
In the he_burn_core.py script, we create a merged library all_lib from core_lib, but
then proceed to still use core_lib for the detailed balance.  This still works in the end
because the _rates dict is shared between the two.  But we should really be operating
on all_lib.  This becomes an issue if we fix this behavior in pynucastro, such that
library addition results in a new library with its own _rates.

The rate ordering here is slightly different now, since the remove rates from the derivation
are now called _derived_removed.


zingale:
This was uncovered by this PR: https://github.com/pynucastro/pynucastro/pull/1112

PR #1852, opened 2025-09-26
add NSE network  compatibility docs

zingale:


zhichen3:
I think it might work if the weak rates are slow so that dYedt is very small then the state reaches NSE whenever it switches to a slightly different Ye? I haven't tested, but technically this test is meant to use for networks without weak rates.

zingale:
okay, that's what I figured.  I will update the comment to reflect this in a bit.  We should provide a way to disable weak rates in pynucastro nets for testing.

zhichen3 (approved):

PR #1856, opened 2025-10-07
fix some compiler warnings if neutrinos are disabled

zingale:
building with USE_NEUTRINOS=FALSE gives warnings of unused vars. This fixes the warnings.


zingale:
oh... there's something off in how we setup those tests, and I initially targeted `main`, so I'll just push to get it going again.

khanakbhargava (approved):

PR #1857, opened 2025-10-09
add check interval for checking nse inside integrator

zhichen3:


zingale:
do we still want this?

zhichen3:
I tested with this pr before and it didn't help with subchandra. I think the current self-consistent NSE-check is reasonably optimized, and if the check itself outweighs benefit that it brings, then we probably just shouldn't use it. I think this method really shines when we have the quasi-statistical stuff hooked in where we work in a reduced system. This is something that I will come back to once I at least have a rough draft for my xrb paper.

At the meantime, I think a traditional temperature check might work now. It didn't work well before since there were a discrepancy between NSE-solution and the steady state solution. But now its fixed. I tested the simple temperature check awhile back with the `Detonation` problem and it worked really nicely -- not sure if I shared the plot.  I was gonna test it with subchandra, but somehow got distracted. I should revisit this...

zingale:
okay, then why don't you close this.

PR #1860, opened 2025-10-17
add an adaptive-NSE network with Fe-group nuclei

zingale:


networks/he-burn/ase-iron/README.md
```diff
@@ -0,0 +1,19 @@
+# `ase`
```
zhichen3: I think this is just a duplicate from ase/README?

PR #1867, opened 2025-10-22
squeeze some more space out of VODE's memory footprint

zingale:


zingale:
microphysics tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2025-10-22-001/index.html

zingale:
this actually results in a 1% slowdown on Frontier (for `flame_wave`)

zingale:
changing the type of the pivot array makes things worse.  If I remove that, the remainder of the changes are mostly just making things const.

zingale:
this still causes answer diffs on Frontier.  Not sure where the issue is

zingale:
gonna close this.  Never seemed to be worth it.

PR #1869, opened 2025-10-24
[WIP] multizone CPU/GPU test harness

BenWibking:
This problem is designed to go through (nearly) the full range of densities/temperatures we encounter a hydro sim without doing hydro.

Prints relative L_1, L_2, and L_inf error norms of the GPU solution wrt the CPU solution.
(All error norms are relative to the CPU solution norm.)


BenWibking:
rocm/6.3.1:
```
starting the multi-zone burn...
CPU multi-zone burn complete; zones failed = 0
CPU multi-zone temperature range: [3032.437486, 3033.069765]
CPU multi-zone retries applied: 1
GPU multi-zone burn complete; zones failed = 0
GPU multi-zone temperature range: [3033.179406, 3033.306148]
GPU multi-zone retries applied: 39
zone 0 T mismatch: CPU = 3032.689294, GPU = 3033.210804
```
I need to compute a better norm to compare them, but it looks like 6.4 is bad for us as well :/

BenWibking:
Just for future reference, @markkrumholz suggests (via Slack):

> I would think that a good norm might be something like the max of the relative abundances for all species except the very low abundance ones; the very low abundance ones a very sensitive to numerical details. So if we define nmax as the abundance of the most abundant species, a good norm might be rel_err = | nX,CPU - nX,GPU | / (nX,CPU + 1e-10 * nmax). Might need to experiment a bit with the numerical factor, but that's basically the norm I use in despotic to determine when a chemistry network has reached steady state.

BenWibking:
This is obsolete. New test harness is on a separate branch.

PR #1870, opened 2025-10-24
update nets for pynucastro changes

zingale:
this captures a lot of optimizations and syncs us up with pynucastro 2.8.0


zingale:
this seems to give a 2% speed up in test_react on groot.  Not much difference seen on Frontier in a full simulation.

PR #1872, opened 2025-10-25
add loop unrolling to MathArray2D

zingale:
this gets about 1% performance boost for flame_wave on Frontier


zingale:
this seems okay now.  We can only put the unrolling into the zero() function without changing the answer on CUDA.

zhichen3 (approved):

PR #1878, opened 2025-11-07
More input species

melilly:
Decided to go the quick route for this problem. 

For future reference, the better, more general solution is: modify the input scheme so Microphysics unit tests can take, as input, the name of a file that details the composition, especially useful for large networks.


zingale:
if you do this, then you also need to update `Microphysics/unit_tests/_parameters`

PR #1879, opened 2025-11-07
regenerate pynucastro nets with caching + other opts

zingale:
this now caches the index search into tabulated rates and partition
function arrays.  It also skips the partition function interpolation
if we know we are below the temperature where it is > 1.  Finally
for tabulated rates, we no longer interpolate quantities that are unused.


zingale:
tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2025-11-08-003/index.html

zhichen3 (approved):

PR #1891, opened 2025-12-03
add a unit test for the cubic interpolation in pynucastro nets

zingale:


zingale:
depends on https://github.com/pynucastro/pynucastro/pull/1191

PR #1893, opened 2025-12-09
eliminate an exp at the low end of screening

zingale:
(found this while testing profiling for our hackathon on Thurs)


zhichen3 (approved):

zingale:
I actually want to think about this some more, since it gives some diffs on GPUs:
http://groot.astro.sunysb.edu/Microphysics/test-suite/gpu/2025-12-09-001/index.html

PR #1899, opened 2025-12-14
some cleaning of screen5

zingale:
this will change answers to roundoff because we do sqrt(sqrt(x)) instead of pow(x, 0.25)


zingale:
here are the tests:
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2025-12-16-001/index.html

if you look at the screening test, you can see that the sqrt(sqrt(x)) change introduces roundoff diffs.  

http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2025-12-16-001/test_screening_C.html

For a few tests, the burning takes an additional few steps, giving seemingly large diffs in the output.  But I think that the diffs are okay.

We might be able to write our own qdrt() function that does the automatic differentiation more efficiently.

zingale:
okay, this is good now.  Here are updated tests:
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2025-12-31-001/index.html

it's best to look at the `test_rhs_*` versions, since those show that this is only roundoff diff

PR #1903, opened 2025-12-17
some cleaning of apply_electrons

zingale:
remove some data that was used for derivatives of degeneracy parameter
also reuse some arrays (sit[]) by scoping where they are filled and used.


zingale:
regression tests pass:
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2025-12-17-001/index.html

PR #1906, opened 2025-12-22
regenerate networks with the new derived rate functionality

zingale:


zingale:
updated with https://github.com/pynucastro/pynucastro/pull/1217

PR #1910, opened 2026-01-07
some more templating of apply_electrons

zingale:
don't compute the electron number density or degeneracy parameter unless needed. Also block off composition derivatives.


zingale:
Microphysics tests: http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2026-01-07-002/index.html

zhichen3 (approved):

PR #1911, opened 2026-01-07
add a utility to find xrb ignition condition following "settle"

zhichen3:
Basically follows: https://github.com/andrewcumming/settle, but use our version of screening / conductivity routines.


zhichen3:
I think this is working. I checked between the solution given by the `settle` code, and they're in the same ball park -- noting that different screening and conductivity routine is used.

zingale (approved):

PR #1917, opened 2026-01-14
allow burn_cell_sdc to take rhoX as input

zingale:
also increase precision in the parse script
I am trying to be able to reliably reproduce failures


zhichen3 (approved):

zingale:
this has been updated so the storage is consistent with the changes to the script

PR #1919, opened 2026-01-15
[WIP] disable integrator.do_species_clip by default

zingale:
it seems to cause more problems then it fixes


zingale:
The iso7 FPEs are fixed if we use the option in #1921

PR #1921, opened 2026-01-15
VODE: bail out of the nonlinear solve if we predict negative mass fractions

zingale:
Also disable `integrator.do_species_clip` by default.  Clipping the integation state seems to do more
harm than good.


zingale:
regression tests with this new option enabled:
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/2026-01-15-001/index.html

zingale:
this is ready for review.

PR #1932, opened 2026-01-25
NSE_NET check update

zhichen3:
Do some updates for the NSE_NET check. The primary change is to do all the rate filtering when writing out the rate index using pynucastro instead of doing it at runtime. 

I made `in_single_group()` strictly to check for ONE group. Previously I optionally allow two groups: one for the light-isotope-group and the other one. If there are no neutron in the net, the definition for NSE configuration is even looser, where I only check for nuclei after si28 and see if they form a separate group for odd Z and even Z.  I don't think this is correct because when we have configurations like this, this means that the integrated mass abundance will NOT be the same as the NSE abundance. In other words, if the network is intrinsically compatible with NSE, then this should not be needed.

### DIFF
--- a/.github/workflows/check-ifdefs.yml
+++ b/.github/workflows/check-ifdefs.yml
@@ -15,13 +15,11 @@ jobs:
 
     steps:
       - uses: actions/checkout@v6
-        with:
-          fetch-depth: 0
 
       - name: Setup Python
         uses: actions/setup-python@v6
         with:
-          python-version: '3.11'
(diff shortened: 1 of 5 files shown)


zhichen3:
One of the "edge" case in `make_table` for NSE_NET changed. But I think its fine. Now in order to be considered as NSE, I now strictly allow ONE single group in the end. Previously I also optionally allowed the TWO groups -- LIG and non-LIG group.

Issue #1946, opened 2026-03-03
NSE linear interpolation can index past upper bound


BenWibking:
# Linear NSE table interpolation can index past table bounds at upper edge

## Summary
In `nse_tabular/nse_table.H`, `nse_interp()` clamps `logrho/logT/Ye` to table maxima, then computes linear interpolation indices (`ir1`, `it1`, `ic1`) and calls `trilinear()`, which accesses `index+1`. At the upper edge, indices can become the last table point, causing out-of-range `+1` access.

## Severity
**High**

## Affected Code
- `nse_tabular/nse_table.H`
  - index calculation in linear mode around lines 429-431
  - neighbor access in `trilinear()` around lines 179-186

## Why This Is a Bug
If `rholog == logrho_max` (similarly for `T`/`Ye`), `nse_get_*_index()` can return the final grid index. `trilinear()` then reads `ir1+1`/`it1+1`/`ic1+1`, which exceeds valid table bounds.

## Proposed Patch
Clamp linear-mode base indices to `[1, N-1]` before calling `trilinear()`:

```diff
--- a/nse_tabular/nse_table.H
+++ b/nse_tabular/nse_table.H
@@
-        int ir1 = nse_get_logrho_index(rholog);
-        int it1 = nse_get_logT_index(tlog);
-        int ic1 = nse_get_ye_index(yet);
+        int ir1 = nse_get_logrho_index(rholog);
+        int it1 = nse_get_logT_index(tlog);
+        int ic1 = nse_get_ye_index(yet);
+
+        ir1 = amrex::Clamp(ir1, 1, nse_table_size::nden - 1);
+        it1 = amrex::Clamp(it1, 1, nse_table_size::ntemp - 1);
+        ic1 = amrex::Clamp(ic1, 1, nse_table_size::nye - 1);
```

## Validation
- Evaluate `nse_interp()` at all table-edge corners (min/max in each dimension) with `nse_table_interp_linear=1`.
- Confirm no out-of-bounds and interpolation continuity near upper boundaries.

zingale:
we never hit this because we clamp the input thermodynamic state to the table limits.  So we can't seem to go out of bounds.  I'll add an assert just in case though, as a check (for debug runs).

PR #1951, opened 2026-03-05
add an assert on the NSE table index

zingale:
we should not be going out of bounds because we clamp the input state to the table limits, but this adds an additional check when run in debug mode.


zingale:
closes #1946

zhichen3 (approved):

Issue #1953, opened 2026-03-05
RKC: nstsig update uses wrong precedence, disabling periodic spectral-radius refresh


zingale:
### Summary
`rkclow()` updates `nstsig` using:

```cpp
nstsig = nstsig+1 % 25;
```

Because `%` binds tighter than `+`, this expression is parsed as:

```cpp
nstsig = nstsig + (1 % 25);
```

so `nstsig` increments by one forever and is never reduced modulo 25.

### Why this is a bug
Immediately after this update, code checks `if (nstsig == 0) { newspc = !jacatt; }`, which is intended to periodically force spectral-radius recomputation. With the current expression, `nstsig == 0` is effectively never true after initialization, so that refresh path is disabled.

### Location
- `integration/RKC/rkc.H` (accepted-step block)

### Suggested fix
Use parentheses (or equivalent) for modulo wrapping, e.g.:

```cpp
nstsig = (nstsig + 1) % 25;
```

PR #1955, opened 2026-03-05
fix the logic for when to refresh the spectral radius in RKC

zingale:
The fortran code was:
  nstsig = mod(nstsig+1,25)
but this was converted to C++ as:
  nstsig = nstsig+1 % 25
but % has higher precedence than +, so we were doing this wrong


zingale:
see the original Fortran code here:
https://netlib.org/ode/rkc.f

closes #1953

zhichen3 (approved):

PR #1963, opened 2026-03-12
add a mul_plus_identity function to MathArray

zingale:
this let's us do J -> I - tau J in a single pass


zingale:
this actually slows things down.

PR #1966, opened 2026-03-20
Updated primordial chem network

psharda:
This PR updates the primordial chemistry network to incorporate a couple of additional He reactions that are important at high temperatures, and use the same combination of H2 dissociation rates as Hunter et al. 2023.


BenWibking:
Is this all SymPy-generated changes? Since the tests pass I will not perform a line-by-line review if that's the case.

networks/primordial_chem/actual_rhs.H
```diff
@@ -1,9 +1,6 @@
-//Last updated July 2024
-//Update notes: 1. the original version of this file used std::pow, which we replaced with a combo of std::exp/std::log wherever possible.
-//Update notes: 2. added free-free cooling which was missing from the previous version
+//Last updated March 2026
 //Author: Piyush Sharda (sharda@strw.leidenuniv.nl)
```
BenWibking, line 2: Could you add what version of your code you used to generate this? I think this will help in the future if it has to be regenerated.

BenWibking (changes requested):
If it's all symbolic regenerated stuff, I won't perform a line-by-line review.

I think it would be good to adopt the standard practice of including what version (including git commit) you used to symbolically generate this for future reference.

BenWibking:
This doesn't have to go in this PR, but we should discuss in the near future if we can find a way to avoid calling `pow` in this network. Performance should be much better if we can avoid it.

One option is to precalculate most of the rates and tabulate them, along the lines of JAFF's `export_table` feature: https://github.com/jaff-chemistry/jaff/blob/main/examples/export_table.ipynb

BenWibking:
> This doesn't have to go in this PR, but we should discuss in the near future if we can find a way to avoid calling `pow` in this network. Performance should be much better if we can avoid it.
> 
> One option is to precalculate most of the rates and tabulate them, along the lines of JAFF's `export_table` feature: https://github.com/jaff-chemistry/jaff/blob/main/examples/export_table.ipynb

I forgot we already removed all of the `pow` calls. But there are still several hundred `exp` and `log`. But that can be changed later.

networks/primordial_chem/actual_rhs.H
BenWibking, line 1: It looks like you reintroduce `std::pow` here. In an earlier PR, you replaced all the `pow` calls with `std::exp([...] * std::log([...]))`, which improved performance. Can you do that rewrite here?

BenWibking (changes requested):

BenWibking:
@psharda ping

networks/primordial_chem/actual_rhs.H
psharda: Right. There was a script available within Microphysics to do this, right? I need to check

BenWibking: I thought you wrote a script to do this, but I can't remember the details.

PR #1969, opened 2026-03-26
make species_failure_tolerance a runtime parameter

zingale:
the current value of 0.01 is actually quite large.  If we get a mass fraction of -0.01, we've already gone bad and it is hard to recover.  Seeing species_failure_tolerance closer to atol_spec can really help with the step rejection logic.


zingale:
for one of the massive star runs (aprox19 with tabular NSE), if we change from the default of 0.01 to 1.e-4, then we go from VODE failure after 500,000 steps to just needing 22 steps.

zingale:
regression tests pass:
http://groot.astro.sunysb.edu/Microphysics/test-suite/gfortran/

zhichen3 (approved):

PR #1975, opened 2026-03-31
Feature/radiation vode

James471:
The PR adds radiation number densities and fluxes and reduced speed of light in burn state. It also makes changes to the VODE integrator to handle photochemical reactions. The PR also fixes a bug that causes the last character of eos file to not be read if the files doesn't end with a newline character.


BenWibking:
@zingale This is something that is critical for our simulation in the second half of this year. We'd like to merge this in the next week or two if possible. Let us know if there's any additional context or review guidance we can provide.

A follow-up PR will make analogous changes to the Rosenbrock integrator to add support for radiation variables.

chongchonghe:
A description of the physics for the photoionization network is here: https://github.com/quokka-astro/quokka/issues/1800

PR #1976, opened 2026-04-02
fix some logic for initial_timestep

zingale:
we were clamping the wrong value


zhichen3:
i can approve this if you fix the burn cell CI

zingale:
I think I've fixed them

zhichen3 (approved):

PR #1979, opened 2026-04-23
add a simple unit test that just evaluates a tabular weak rate

zingale:
this requires a pynucastro net.
This is for debugging the interpolation.


zhichen3 (approved):

zingale:
I need to add this to the docs and then I'll request as re-review

zingale:
okay, docs are updated

zhichen3 (approved):

PR #1994, opened 2026-05-11
Add Rosenbrock integrator

BenWibking:
## Summary
Adds a new `integration/Rosenbrock` integrator implementation with runtime-selectable tableaus and timestep controllers.

The integrator supports both Strang and SDC setup paths through separate `actual_integrator` headers, uses the existing RHS/Jacobian/linear-solve interfaces, and exposes runtime parameters for tableau selection, rejection behavior, and controller tuning.

*List of methods:*
  - Rodas5P (**default, works best on all tests**; https://doi.org/10.1007/s10543-023-00967-x)
  - Rodas4P (https://pub.h-brs.de/frontdoor/index/index/docId/1548)
  - Rodas3P (https://github.com/hbrs-cse/RosenbrockMethods/blob/main/paper/JuliaPaper.pdf)
  - ROS2S (https://doi.org/10.1016/j.apnum.2012.08.001.)
  - ROS2 (https://scixplorer.org/abs/1999SJSC...20.1456V/abstract)
  - Rosenbrock-Euler/YASS (Khoklov, unpublished)

[Note: The "P" suffix refers to methods developed to satisfy the stiff accuracy conditions derived by [Prothero & Robinson (1974)](https://ww2.ams.org/journals/mcom/1974-28-125/S0025-5718-1974-0331793-2/S0025-5718-1974-0331793-2.pdf). ROS2S also satisfies these conditions.]

*List of timestep controllers:*
  - H211b error-history timestep control (**default**; https://doi.org/10.1145/641876.641877)
  - YASS concentration-change timestep control

## Related PRs
Depends on https://github.com/AMReX-Astro/Microphysics/pull/1595.


BenWibking:
@zingale Here's the draft PR. This does not have all the performance optimizations that I tried.

BenWibking:
This version wins by a few percent on CPU [Mac laptop] over VODE for primordial_chem when building both with Clang (which is much faster than gcc on CPU for these problems).

The performance delta is significantly larger on a Hopper GPU (in favor of Rosenbrock).

zingale:
this gets an FPE if I build and run `test_react` as:

```
make INTEGRATOR_DIR=ROS2S DEBUG=TRUE
./main3d.gnu.DEBUG.ex inputs_aprox13  amrex.fpe_trap_invalid=1
```

zingale:
it seems to be that it is generating really negative mass fractions and that's causing the failure, so we probably need more step rejection logic.

BenWibking:
The docs fail due to this broken link (or Zenodo not accepting connections from the CI?)
```
(          citing: line    2) broken    https://doi.org/10.5281/zenodo.2620544 - 403 Client Error: Forbidden for url: https://zenodo.org/doi/10.5281/zenodo.2620544
```

zingale (approved):

Issue #2007, opened 2026-05-29
add fallback to numerical Jacobian if analytic Jacobian is NAN


BenWibking:
This can happen due to numerical underflow in the analytic Jacobian, especially at low temperatures (where rate derivatives often underflow).

PR #2008, opened 2026-05-30
add fallback to finite diff Jacobian

BenWibking:
In the Rosenbrock integrator, add a fall back to the FD Jacobian if the analytic Jacobian produces a NaN or Inf value.

Fixes https://github.com/AMReX-Astro/Microphysics/issues/2007.


BenWibking:
@zingale Fixed to use isnan/isinf.

zingale:
Shouldn't it use amrex::Gpu::isnan()

BenWibking:
> Shouldn't it use amrex::Gpu::isnan()

Why? We use `std::isnan` in my places in device kernels in our code. I haven't seen any problems.

zingale:
it looks like it only matters with SYCL

zingale:
can you fix the tabs?

BenWibking:
@zingale tabs fixed.

zingale (approved):

PR #2009, opened 2026-05-31
update ase network  to use __device__

zingale:
this is needed for HIP
also need to switch the ase CUDA test to use `test_sdc` instead, since that is GPU-enabled.  `nse_net_cell` does not have any GPU launch.


khanakbhargava:
CUDA test is failing

khanakbhargava (approved):

PR #2013, opened 2026-06-01
fix three-way loop bug

chongchonghe:


BenWibking:
Since you created a new PR to James' PR, this PR can be closed, right?

PR #2014, opened 2026-06-07
Remove YASS

BenWibking:
Removes the YASS method from `integrators/Rosenbrock`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+
 # latex intermediates
 *.aux
 *.log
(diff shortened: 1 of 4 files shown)


zingale:
for some reason the checks aren't running.  I think we need to push something here to retrigger them.

BenWibking:
ok, let's see if that worked.

BenWibking:
@zingale It looks like everything passed.

zingale (approved):

PR #2015, opened 2026-06-08
Remove Rosenbrock-Euler and Ros2 methods

BenWibking:
Removes the Rosenbrock-Euler and Ros2 methods. These are not performance-competitive for any networks.

The remaining Rosenbrock methods all have both stiff accuracy and stiffly accurate embedded error estimators (which Rosenbrock-Euler and Ros2 do not have).

### DIFF
--- a/Docs/source/ode_integrators.rst
+++ b/Docs/source/ode_integrators.rst
@@ -70,10 +70,6 @@ Presently, allowed integrators are:
 
   * ``3`` : ROS2S method, a 2nd order, stiff-accurate method :cite:`ros2s`.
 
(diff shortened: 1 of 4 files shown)


BenWibking:
@zingale This just removes the methods that were too slow to be useful for any network either of us have tested.

zingale:
can you update the docs?
https://amrex-astro.github.io/Microphysics/docs/ode_integrators.html

BenWibking:
> can you update the docs? https://amrex-astro.github.io/Microphysics/docs/ode_integrators.html

Fixed.

zingale (approved):

PR #2016, opened 2026-06-08
[WIP] Reduce Rosenbrock memory usage

BenWibking:
Reduce memory usage for Rodas5P by re-using work arrays in a non-trivial way:

* While the final update does require all previous stages, we can save memory by only saving the linear combinations of previous stages that are re-used by later stages. This reduces the work array usage to the high-water mark of the number of linear combinations of previous stages that are needed at any given time.
* The total number of state arrays needed in `rosenbrock_t` is reduced 9 arrays to 5 arrays (including the input/output state arrays).

Unfortunately, the code to do this is quite ugly. This PR is still marked "WIP" while I figure out how to make it simpler.

This only reduces memory usage for Rodas5P.

### DIFF
--- a/integration/Rosenbrock/_parameters
+++ b/integration/Rosenbrock/_parameters
@@ -18,3 +18,4 @@ h211b_k                      real         2.5
 h211b_fac_min                real         0.2
 h211b_fac_max                real         6.0
 h211b_reduction_fac          real         0.5
+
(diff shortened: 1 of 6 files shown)


BenWibking:
This changes the answers due to floating-point reassociativity differences. Trace species below atol_spec change more significantly.

Issue #2019, opened 2026-06-20
add state checks for SDC to Rosenbrock


zingale:
The Rosenbrock functions `valid_integtrator_state` and `species_change_valid` are not implemented for SDC.

PR #2027, opened 2026-06-30
Rosenbrock: add step-rejection logic for SDC

zingale:


zingale:
closes #2019

zingale:
it looks like we are using `species_failure_tolerance` in VODE instead of `atol_spec` for the constraint.  So maybe we should test that.

PR #2029, opened 2026-07-17
add a big ASE network

zingale:
this has 61 nuclei


zhichen3 (approved):

zhichen3:
actually, make sure this is generated with the latest pynucastro version -- in case there are any significant changes.

zingale:
updated

PR #2038, opened 2026-08-04
Initial tweaks for JOSS submission ahead of publication

warrickball:
This is an initial set of editorial tweaks to the JOSS paper following the [successful review](https://github.com/openjournals/joss-reviews/issues/10422). Most of these are to the citation style and a few very minor tweaks to the text. 

Note that I've changed the "no affiliation" to "Independent Researcher, USA", as in [our example paper](https://joss.readthedocs.io/en/latest/example_paper.html) but that assumes that the unaffiliated contributors are in the USA (or would like that to be their national affiliation). If that's wrong, please update it accordingly, even if it means those two authors have separate "Independent Researcher" affiliations to allow different countries.

I'll also flag that you haven't included any ORCIDs for the authors. We recommend it but don't require it.

If you're happy with the changes and merge them (or selectively apply some yourself), I'll launch another test run of our publication pipeline to double-check I haven't missed anything else.


zingale (approved):

zingale:
thanks.  I'll add the ORCiD now.  I am not sure why I left them out -- they are in the `.zenodo.json`.  I need to look up the syntax.

PR #2087, opened 2026-09-09
make ForwardEuler respect subtract_initial_energy

zingale:
now we can run a burn_cell test without crashing


integration/ForwardEuler/actual_integrator.H
zhichen3, line 176: we already have `e_in` here

zhichen3: this print also needs to be changed accordingly

zingale: indeed.  I missed that.

zingale: fixed

zhichen3 (approved):

PR #2092, opened 2026-09-12
[WIP] Use monotonic cubic Hermite interpolation for the NSE table

zingale:
This addresses an issue where even if the data were all positive, we could get negative values in the interpolation due to undershoots in the existing cubic interpolation.

There is one downside though -- because the limiting is nonlinear, it no longer guarantees that sum{X_i} = 1.


zingale:
I did a small run of `massive_star` with these changes and did not notice much difference with the old method.  But this should be safer overall.

